PR #45, opened 2014-07-21
uon merge master

drjosephbaxter:
Hi Dan,

UoN changes for Master branch. We are not on planning on moving to 2.7 for a while (we move to 2.6 in August) but I have done a quick sanity check and it seems fine.

Joe


danmarsden:
@drjosephbaxter I've recently added automated travis CI checks here on github which is really cool :-) - just looks like your old attendance export file is ok test doesn't work due to the dependency on behat_download - I've recently fixed up the other behat tests to make sure they are running but I might end up removing the file is ok check so that the build is successful - let me know if you might have any time to remove the dependancy or adjust the automated travis checks so that it passes.
automated tests can be seen here:
https://travis-ci.org/danmarsden/moodle-mod_attendance

PR #46, opened 2014-07-31
Bug when trying to add a new session

ptramonti:
I found a little bug in the Add new session page. When “Multiple sessions” is selected and the user select days of week that are not in the date interval selected above, once “Save” is pressed the page returns a “Programming error”. I fixed it.


danmarsden:
I've just taken another look at this and I can't see any programming errors when you select days that don't exist in the date interval range - it's possible this has been fixed in a different patch. I don't think we need this unless I'm missing something else.

ptramonti:
Hi Dan,
thank you for the response!
I've just pulled the master and checked if a patch was already applied to fix this issue, but i've encountered the same error.
I attach some screenshot to describe better the issue:

1. Step to reproduce the issue (both 28th and 29th January are not Friday)
schermata del 2015-01-28 10 31 15


2. What happens, after submitting, now
schermata del 2015-01-28 10 31 22


3. What happens, after submitting, with the fix
schermata del 2015-01-28 10 40 12

danmarsden:
ah - when I tested I used a date range that included one of the days.... Will re-test and merge - thanks!

danmarsden:
I've rebased those commits into a single commit and pushed it in - thanks!

PR #47, opened 2014-07-31
New feature for Statuses

ptramonti:
New feature for statuses. It gives the possibility to specify (with a checkbox in add/edit statuses page) for a status, e.g., saved as “Justified”, to not combine to the total of presences/absences. This feature is related to the percentage one. If a student must have a minimum percentage of 60% to access final exam, and in a total of six lessons his report is P-A-P-J-P-A, if status J is like A, without the new feature, the percentage would be 3P/6 = 50%, in the other case (with feature and Grade set to 0) would be 3P/5 = 60%.
In addition, this feature could be useful also to gain Grades in extra lessons. If this feature is selected for a status (E) with e.g. Grade set to 1, attending a lesson with status E means make own percentage grows.

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -70,6 +70,7 @@
         <FIELD NAME="grade" TYPE="int" LENGTH="3" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
         <FIELD NAME="visible" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="1" SEQUENCE="false"/>
         <FIELD NAME="deleted" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
+        <FIELD NAME="calc_total" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="0" SEQUENCE="false" COMMENT="Consider absence as justified or not"/>
       </FIELDS>
       <KEYS>
         <KEY NAME="primary" TYPE="primary" FIELDS="id" COMMENT="Primary key for attendance_settings"/>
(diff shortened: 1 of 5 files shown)


danmarsden:
Sorry for the delay here, I think this needs a bit more thought as it could add a further level of complexity to the interface. If you could add some screenshots and more detailed info on how this works it would be great! - unfortunately this pull request won't merge anymore so will also require rebasing.

ptramonti:
I try to explain better the feature.

In 10 sessions (e.g), a student could be Absent or Present. So, to see the attendance in the course of that student, the calculation could be total_presences/total_sessions, where total_sessions = total_presences+total_absences.

In such situation, we could not "justify" an absence. So I introduced a field in the Add/Edit Statuses tab. A status could be flagged as "Extra", so that it could improve the total_presences value but not the total_sessions.

For example in this screenshot...
schermata del 2015-01-29 09 46 56
... the student Test2 has 2/2 as a grade, even though sessions are 3. That's beacuse he has 2(P)+1(E), and E is configured as...
schermata del 2015-01-29 09 47 02
... with "Extra" field flagged TRUE and "Grade" set to 0.


This improvement could be also useful to give some "added value" to a student in a session in which was very brilliant.
So we can add a status "Good!" with "Extra" flagged TRUE and Grade set to 1.








schermata del 2015-01-29 09 48 30


schermata del 2015-01-29 09 48 23

So student Test3, even though an Absence not Excused, will have 100% of attendance beacuse of a Good!.

Is it maybe too specific? Or unuseful?

danmarsden:
Thanks for providing that extra detail!

My initial feelings are that it seems too specific to your use-case and the complexity that this introduces into the interface for teachers makes it hard to justify. If you think there's likely to be a large number of people who really want this in the main code-base I'd suggest posting in the Attendance forum explaining your feature.
https://moodle.org/mod/forum/view.php?id=741

If you get general support in the forums I could still be convinced to add this into the core code. 

Thanks!

danmarsden:
closing this pull request for the following reasons:
1) the pull request no longer merges and requires a rebase
2) the change introduces complexity to the interface that I think needs improvement - particularly around heading names, descriptions etc.
3) the change appears to be specific to a particular use case and the complexity added appears all the time and may be better implemented as something than can be turned on/off. (should probably be off by default)

I'm happy to review the change again if those can be improved - thanks for sharing!

PR #48, opened 2014-07-31
Percentage in session reports

ptramonti:
Percentage of attendance in session reports, with the possibility to specify a minimum value to highlight who achieved it and who didn’t.

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -10,6 +10,8 @@
         <FIELD NAME="course" TYPE="int" LENGTH="10" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
         <FIELD NAME="name" TYPE="char" LENGTH="255" NOTNULL="false" SEQUENCE="false"/>
         <FIELD NAME="grade" TYPE="int" LENGTH="10" NOTNULL="true" DEFAULT="100" SEQUENCE="false" COMMENT="This is maximum grade for instance"/>
+        <FIELD NAME="min_attendance" TYPE="int" LENGTH="3" NOTNULL="true" DEFAULT="65" SEQUENCE="false" COMMENT="Minimum attendance percentage to access final exam"/>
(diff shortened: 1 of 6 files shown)


danmarsden:
Thanks for the patch, I think we need to improve the reports a bit and possibly just add the percentage without requiring a new setting to turn it on/off. Also I'm not sure about the "min_percentage" feature - it would be good if you could flesh this out a bit more and provide some screenshots.

ptramonti:
Hi Dan,
some courses needs from the students a minimum attendance to take (e.g.) the final exam or (e.g. 2) achieve the course. This implementation works in this direction, providing the teacher a quick view of students enabled to take the exam/which passed the course or not.
About the setting on/off, some teachers want the percentage and others don't, but it could be also shown by default without the need to switch the on/off feature.
Here are some screenshots about how this implementation works.

1. show_percentage -> TRUE && min_attendance SET (>0) 
schermata del 2015-01-28 12 01 39
schermata del 2015-01-28 12 19 42


2. show_percentage -> FALSE && min_attendance SET (>0)
schermata del 2015-01-28 12 01 48 
schermata del 2015-01-28 12 19 58


3. show_percentage -> TRUE && min_attendance !SET (=0) 
schermata del 2015-01-28 12 02 37
schermata del 2015-01-28 12 19 30

danmarsden:
I wonder if we could implement this function using the standard_grading_coursemodule_elements which includes a gradepass setting.

ptramonti:
Hi Dan,
I followed your tip and checked out the standard_grading_coursemodule_elements() function.
In this way the percentage is shown in the gradebook section instead of attendance > report (and requires much less code lines to write!).
I see also that it works both in 2.87 and 2.9.1 version of Moodle, but it shows the grade green/red (depending on the grade is greater/lower than the gradepass value) only in 2.9.1 version. Anyway, it's not a big problem.
Then I close this pull request and open a new one with the brand-new code edits.

ptramonti:
New pull request: https://github.com/danmarsden/moodle-mod_attendance/pull/130

PR #68, opened 2015-01-29
Translation pt br

antonio-c-mariani:
Small changes related to translation and translation to portuguese (pt_br)


danmarsden:
Thanks - all Translations are stored in lang.moodle.org rather than individual plugin git repositories, if you could remove the portugese translation here and submit it to lang.moodle.org that would be great.

antonio-c-mariani:
Ok, I did that. But there are two or three small changes that must be done in your code related do translation. Have you seen that in the previous pull request? Is it the case of doing a new pull request only with this changes?

danmarsden:
yeah - if you file a separate pull request with just that change it would be great. Thanks.

PR #77, opened 2015-02-13
Fix #74 - crud and objecttable params in /classes/event/ need fixing

### LINKED ISSUE #74: crud and objecttable params in /classes/event/ need fixing

barrysspace:
This fixes issue #74


danmarsden:
makes complete sense to me! - thanks for the patch!

PR #78, opened 2015-02-13
Fix #73 - Debug warnings when adding session for "Separate groups" if no...

### LINKED ISSUE #73: Debug warnings when adding session for "Separate groups" if no groups defined
Steps to reproduce:
1. Create attendance instance with "Separate groups" group mode in a course with no groups defined
2. Add a session
3. Warnings are displayed as below:

add_session_error

barrysspace:
This fixes issue #73


danmarsden:
thanks!

PR #79, opened 2015-02-13
Fix #75 add_record_snapshot() errors

### LINKED ISSUE #75: add_record_snapshot() gives debug messages on various pages
Steps to reproduce:

1. Turn Developer debugging on
2. Add/update a session, take attendance, add/update statuses, etc
3. Messages will appear similar to screenshot below:

add_record_snapshot_debug_messages

There is also no "status_removed" event.

barrysspace:
This fixes issue #75 and adds a new status_removed event

### DIFF
--- a/classes/event/status_removed.php
+++ b/classes/event/status_removed.php
@@ -1,0 +1,100 @@
+<?php
+// This file is part of the Attendance module for Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks Barry, you appear to be removing a few calls to $event->add_record_snapshot('attendance', $this); - although I agree they probably didn't need to be there in the first place I wonder if this may cause problems for people that are using these events already?

danmarsden:
also - it would have been helpful if the commit to add a new event was seperate from the bug fixing of the existing events but don't worry about that now!

barrysspace:
Hi Dan, you're welcome. To do $event->add_record_snapshot('attendance', $this) properly we would have had to make a database call to get the attendance table record in it's proper form (since $this is not an instanceof \stdClass) or we would need to assemble a new  \stdClass object with all the fields populated.  So by removing this useless line (nothing was changed in the attendance table anyway) I think we are staying with the purpose of add_record_snapshot which is to increase performance overall.

If anyone was listening for these events and wanted to retrieve a record of the attendance table they would need to use get_record_snapshot() to retrieve the cached data.  get_record_snapshot() does a db query if no data is found in the cache.  So this way we only do the hard work (db query) for those who really want it.  What do you think?

danmarsden:
You've convinced me - merged it in. I would like to rewrite the way $this is used within attendance at some point - it's not always very efficient and passes a lot of data around in memory when it isn't always needed.

Thanks!

barrysspace:
That's great, thanks Dan.

PR #90, opened 2015-05-25
adding take_attendance and add_sessions webservices

danielneis:
Hello, Dan

I've added a webservice function to register attendance logs,
based on what is done by "take_from_form_data".

I will add at least one more to create sessions soon =)

Hope that you like =)

Kind regards,
Daniel

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -27,6 +27,7 @@ defined('MOODLE_INTERNAL') || die();
 global $CFG;
 require_once($CFG->libdir . '/gradelib.php');
 require_once(dirname(__FILE__).'/renderhelpers.php');
+require_once($CFG->libdir . '/filelib.php');
 
 define('ATT_VIEW_DAYS', 1);
 define('ATT_VIEW_WEEKS', 2);
@@ -82,21 +83,29 @@ class attendance_permissions {
         require_capability('mod/attendance:viewreports', $this->context);
     }
 
-    public function can_take() {
+    public function can_take($user = null) {
+        if (is_null($user)) {
+            global $USER;
+            $user = $USER->id;
+        }
         if (is_null($this->cantake)) {
-            $this->cantake = has_capability('mod/attendance:takeattendances', $this->context);
+            $this->cantake = has_capability('mod/attendance:takeattendances', $this->context, $user);
         }
 
         return $this->cantake;
     }
 
-    public function can_take_session($groupid) {
-        if (!$this->can_take()) {
-            return false;
+    public function can_take_session($groupid, $user = null) {
+        if (is_null($user)) {
+            global $USER;
+            $user = $USER->id;
+        }
+        if (!$this->can_take($user)) {
+            return true;
         }
 
         if ($groupid == attendance::SESSION_COMMON
-            || $this->can_access_all_groups()
+            || $this->can_access_all_groups($user)
             || array_key_exists($groupid, groups_get_activity_allowed_groups($this->cm))) {
             return true;
         }
@@ -123,7 +132,7 @@ class attendance_permissions {
     public function require_manage_capability() {
         require_capability('mod/attendance:manageattendances', $this->context);
     }
-    
+
     // Check to see if the user can manage temporary users.
     public function can_managetemp() {
         if (is_null($this->canmanagetemp)) {
@@ -168,9 +177,13 @@ class attendance_permissions {
         return $this->canbelisted;
     }
 
-    public function can_access_all_groups() {
+    public function can_access_all_groups($user = null) {
+        if (is_null($user)) {
+            global $USER;
+            $user = $USER->id;
+        }
         if (is_null($this->canaccessallgroups)) {
-            $this->canaccessallgroups = has_capability('moodle/site:accessallgroups', $this->context);
+            $this->canaccessallgroups = has_capability('moodle/site:accessallgroups', $this->context, $user);
         }
 
         return $this->canaccessallgroups;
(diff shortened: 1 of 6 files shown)


danmarsden:
Thanks Daniel - I think I'd prefer to create a new function that is called by take_from_form_data and the new webservices functions rather than duplicate the code in both places - would you mind moving that into a function and calling it in both take_from_form_data() and your new webservices function?

Thanks!

danielneis:
Hello, Dan

to create this new function, I would like to change the way the parameters are passed to it.
Form data is a little "messy" and that substr call is not necessary if we change just a little
the name of the fields on the form, they can be stored/passed just like the new function I've created.
Are you ok with this too?

Kind regards,
Daniel

danmarsden:
makes sense to me - there's a lot of restructuring I'd like to do at some point in the code, any imrpovements are appreciated!

danielneis:
Hello, Dan

I've refactored the code and now we have a function that is called
both by webservice and by take.php and also have better structured form data =)

Kind regards,
Daniel

danielneis:
Hello, I've added the "add_sessions" webservices too =)

locallib.php
```diff
@@ -961,6 +954,14 @@ public function take_from_form_data($formdata) {
         $event->add_record_snapshot('attendance_sessions', $session);
         $event->trigger();
 
+    }
+
+    function redirect_after_take_from_form($formdata) {
```
danmarsden: is $formdata used in this function? - if not we shouldn't include it as a param.

locallib.php
```diff
@@ -903,46 +903,39 @@ public function take_from_student($mformdata) {
         return true;
     }
 
-    public function take_from_form_data($formdata) {
+    public function take_attendance($formdata) {
         global $DB, $USER;
-        // TODO: WARNING - $formdata is unclean - comes from direct $_POST - ideally needs a rewrite but we do some cleaning below.
+
+        if (isset($formdata->takenby) && !empty($formdata->takenby)) {
```
danmarsden: This makes me a little uncomfortable, it would be better if we could pass $takenby as a param in the function and use $USER->id if it is blank rather than use it within $formdata (the use of $formdata really needs a rewrite anyway)

locallib.php
```diff
@@ -903,46 +903,39 @@ public function take_from_student($mformdata) {
...
-        }
-
         $dbsesslog = $this->get_session_log($this->pageparams->sessionid);
-        foreach ($sesslog as $log) {
-            if ($log->statusid) {
-                if (array_key_exists($log->studentid, $dbsesslog)) {
-                    $log->id = $dbsesslog[$log->studentid]->id;
-                    $DB->update_record('attendance_log', $log);
+        foreach ($formdata->users as $log) {
+            $record = new stdClass();
```
danmarsden: being pedantic here.... but I don't personally like using "record" as a var name... maybe $attendancelog instead? :-)

locallib.php
```diff
@@ -903,46 +903,39 @@ public function take_from_student($mformdata) {
...
+            $record->studentid = $log['userid']; // We check is_numeric on this above.
```
danmarsden: looks like you removed the is_numeric cleaning in the patch above - we need to keep this for my sanity or rewrite the $formdata stuff.

locallib.php
```diff
@@ -903,46 +903,39 @@ public function take_from_student($mformdata) {
...
+            $record->statusid = $log['statusid']; // We check is_numeric on this above.
+            $record->statusset = $statuses;
+            $record->remarks = clean_param($log['remarks'], PARAM_TEXT);
+            $record->sessionid = $formdata->sessionid;
```
danmarsden: is this right? - was previously $this->pageparams->sessionid; If we obtain from the formdata should it be cleaned? (I haven't traced the code)

danielneis:
Hello,

I've fixed the code based on your review.
The use of pageparams is not that compatible with webservices,
so i've preferred to pass it with the form/parameters than get it from page,
but now I've added some validation checks =)

Kind regards,
Daniel

danielneis:
Hello,

a co-worker and friend of mine have reviewed the webservice and found that there were some erros preventing it to work.
I've fixed the code and added a simple script that can be used to test.
It is a not a formal test yet, but it is also an usage example.

Kind regards,
Daniel

danmarsden:
Thanks Daniel, 

I haven't had a chance to review this properly yet, will try to sometime next week.

One of the things I don't like about the overall attendance code is the way it hides away the access control checks into global functions that aren't always checked at the very beginning. I think I'd like to remove some of those and just check the capabilities directly. For now it probably makes sense to just add some capability checks into the webservices calls at the top of the functions to make it really clear that access control is being managed within the webservice functions.

Also - you probably don't need to add the extra Curl class as that should already exist in the core code and it's probably best to leave that test file out for now unless you manage to convert to being a unit test.

danielneis:
Hello, Dan

I've converted the tests to unit tests based on assign module.
It's not working 100% yet as i have to create the sessions before test take_attendance,
will work on this soon.

Also, about the capability checks, on the take_attendance external function there is a call to "$att->perm->can_take_session($pageparams->grouptype, $takenby)", that checks the capabilities.
You think it's really necessary to repeat the checks on  the beggining?

Also, as exceptions are thrown if errors occurs, I've simplified the return value of take_attendance removing the error message, this way return only the session id as int value.

Kind regards,
Danie

danmarsden:
ok - leave the capability stuff as-is for now... one day I might rewrite them so it's more obvious when and where the capability checks are being made. Let me know when you're ready for me to take another look.

thanks!

danmarsden:
closing this pull request as it conflicts and wasn't 100% complete as you mention, but please let me know if you finish this off and want me to take another look!

PR #94, opened 2015-05-27
Fix #93 $record object needs an id

### LINKED ISSUE #93: add_record_snapshot() gives debug message in take_from_student()
The $record object needs an id so that it reflects all the fields of the attendance_log table

barrysspace:
When the fix for #75 is merged in we will need this since we will have 
$event->add_record_snapshot('attendance_log', $record);
which requires that the $record object have the same properties as the fields of the attendance_log table

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -878,7 +878,7 @@ class attendance {
             return false;
         }
         else {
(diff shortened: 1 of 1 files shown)


danmarsden:
I've fixed this one slightly differently - thanks!

PR #66, opened 2015-01-27
Adding a summary report

antonio-c-mariani:
This patch adds a summary report that shows some general data about sessions (including percentages). Details about sessions are not shown. It also adds an new setting parameter for the max grade for all instances.

PR #103, opened 2015-06-11
Adding a summary report  (fixes to issue #66)

danielneis:
Hello, Dan

as I work with professor Antonio Carlos Mariani,
we talked about the problems with the summary and i've fixed them.
Now the two columns have different names:
"Attendance percent for sessions taken" and
"Attendance percent for all sessions"
and also we have removed the "maximum grade" option as now Moodle
has CFG->gradepointdefault , so this is being used instead.
Hope that you like.

Kind regards,
Daniel


danmarsden:
Thanks Daniel, it's closer!

the heading I meant "maxgrades" is a column in the summary report, you've removed the lang string from the original patch and now an empty string error is shown on the summary page as the maxgrade string is missing. I'm wondering if there's a better title we can use there as "Maximum grade" doesn't seem clear to me.

danielneis:
Hello, Dan

sorry for the missing string, I've fixed that this way:
About the "Maximum grade", in the settings page there is a setting for "Maximum points" if you are using points; if you are using scales or not using any, this field is disabled.
So I've changed the code to use that same string, that is 'modgrademaxgrade' from 'grades'.

Also, I've reworded the other two columns to a shorter version:
Taken sessions;
All sessions;

Kind regards,
Daniel

danmarsden:
Hi Daniel,

it looks like this current patch adds an extra column to all the other reports and it also doesn't appear to have the right number of header columns.

I'm still not convinced that the new column header names are clear enough... I think the best option would be to add some contextual help icons to each header that allow users to get a full explanation on what each column provides - does that make sense to you?

(sorry about the conflict with version.php - should be easy to fix.)

danielneis:
Closing this since #140 should be adapating it to new code and add other good things =)

PR #104, opened 2015-06-23
Added description and show description on course

ralferlebach:
It would be great if one could add a descrition to this "activity" (like "This is the attendence list for our course. Please check, whether your attendence has been marked correctly.") and to show this optionally within the course.


danmarsden:
Hi there, 

thanks for the suggestion but your code isn't valid or complete and doesn't follow good coding guideline practice. I've created a commit here that implements the same thing as your code to show you how it should be done:
https://github.com/danmarsden/moodle-mod_attendance/commit/3dce3318df0d3b218135e152885ebdc8d30c5589

But - that just adds the fields to the db and doesn't actually display the content from the intro field anywhere in the interface. If you would like to get this feature added it would be great if you took that patch I linked above and finished it off by adding the intro content into the user-interface and re-filing a pull request with the changes.

Thanks!

PR #106, opened 2015-06-25
Fix to make $PAGE->url valid by adding the action parameter.

michael-milette:
Signed-off-by: Michael Milette <michael.milette@tngconsulting.ca>


danmarsden:
great thanks!

PR #121, opened 2015-07-15
Fix issue#120. Show error when no sessions are generated.

barrysspace:


danmarsden:
Closing as fixed using a different method - let me know if you think we should re-open. Thanks!

PR #130, opened 2015-07-21
Adding gradepass feature using standard functions

ptramonti:
Using the Moodle core standard_grading_coursemodule_elements() function to show the grade section in the attendance module settings.
Starting from version 2.9.x this includes the gradepass feature, that permits to show, in gradebook report, the percentage of attendance in green if the grade is greater/equal than the gradepass value, otherwise is shown in red.


danmarsden:
Great - thanks for the patch!

PR #140, opened 2015-08-05
Fix the incorrect grades(points) on reports and other things

antonio-c-mariani:
Hello, Dan

I followed your suggestion and made a series of changes on the attendance mod code in order to fix the problems in the reports. Looking deeply in the code, I tried to organize the counting procedure putting it in only one method: 'att_get_users_points'. This method is used in many different places where the counting was necessary: report, export, and update grades.

I made some other changes that seem interesting like:
- Included a summary report (previously proposed)
- Included a global setting 'defaultperclowgrade', a kind of threshold for the lowgrade report (renamed to "Below n%")
- Changed the concept of 'grades' to 'points' on the reports. The points are always showed even when the attendance has no grades. This seems important to teachers and students because they always want to see the points.

Antonio C. Mariani


danmarsden:
Thanks for the pull request! - looks like quite a few changes there so it might take me a bit of time to review it properly. If you haven't heard from me within a couple of weeks flick me another message to remind me!

antonio-c-mariani:
Ok.

This plugins is very useful to us and the concept of "status set" is very good. But we need to fix the reports and the way the grades are computed before update it to the new version. 

Please let me now if you have any doubts.

BTW, there is a sintax error on the preferences.php of your code.

Antonio C. Mariani

danmarsden:
Thanks for working on this Antonio! - it's always great to have other people contributing to this plugin!

Unfortunately your code doesn't work on Postgres due to errors like this:
" ERROR: column "stg.grade" must appear in the GROUP BY clause or be used in an aggregate function"

Also - modifying the existing function att_get_user_taken_sessions_count() by changing the order of params and adding others could be problematic if people with 3rd party code leverage this function - it would probably be better to use a new function name rather than modifying it. If this was core code we'd need to support the old function name and structure with the new code just in case other people were calling it in their own code but I think it's probably ok to remove the old function and replace it with the newer one.

I would also prefer each individual "feature" or "improvement" to be in a separate commit - (for example the added summary report) as this makes it much easier to review - if you can separate out individual improvements into separate commits it would be really helpful.

it would also be really good for all new functions to use the correct frankenstlye naming - particularly in locallib.php - most of the existing functions don't follow this but moving forward all new functions should.

I haven't checked to see if your code decreases the number of sql queries made to the db when displaying reports - but it would be good if you could check this if you have a chance.

I haven't tested the interface or viewed your actual changes due to the postgres error - hopefully you might have time to improve the patch further and submit a new pull request (thanks for pointing out the issue with preferences.php, I've fixed that up.)

thanks

PR #142, opened 2015-08-23
Some corrections and improvements

antonio-c-mariani:
Hy Dam

Here goes another pull request. As you suggested, the changes were separated in 5 commits. The Postgres error was fixed. As your code is, it's not possible to take attendance for group sessions. Also, some inconsistencies may happen when an group is removed from the course or a student is removed from a group. I didn't see a simple way to avoid these. I made some changes trying to keep the reports consistent just showing messages on these cases.

There are some improvements that is very useful to our teachers because of some characteristics of our classes. I tried to include them as options to the original behavior.

Antonio C. Mariani


danmarsden:
Thanks Antonio - ideally individual bug fixes (group sessions etc) should relate to an issue within the issues list here in the github repo that states the details of the bug/error messages/behaviour etc and then the commit message should reference that bug number.

in an ideal world there would also be an issue in the github tracker here for each individual improvement describing what the new feature does but I can try to determine that based on your commit messages for now.

danmarsden:
Thanks again for working on this Antonio! 

Sorry - we are going to need to have these separated out into individual issues with full details on the changes and pull requests that correspond to the changes.

for example - I'm not clear on why you have changed the code from date_time_selector to date_selector and added a separate session time option - this seems to decrease the usability to me. in that same commit I don't like the way you check the setting and then use a different group name with a different string but replicate the adding of the hours and minutes values. if a different string was required depending on the internal settings, the only thing that should change is the string.

It may be that some of your patches can be integrated as individual items, but I can't merge this particular group of patches as-is sorry. If you could separate it out a bit better into multiple pull requests I can then comment on each individual change and we may be able to integrate some of it?

thanks!

danmarsden:
I did find that the hidden session delete patch worked well though. so I created #143, modified your commit message and merged it into master - looks great thanks!

PR #153, opened 2015-09-11
US date format

Syxton:
m/d/y instead of d.m.y


danmarsden:
Thanks for the patch but the EN_us lang pack is maintained by Moodle's language system on: https://lang.moodle.org/

Only the single "En" lang pack is included in most plugin repositories.

I'm not sure why the attendance plugin maintains it's own lang strings for dates either... We should probably switch them out and use the standard core lang strings.

PR #169, opened 2015-10-02
Fix #155 properly. Undo regression caused by moving $statuses definition

### LINKED ISSUE #155: Hiding statuses on Settings page removes statuses form the list
Steps to reproduce.
On the "Settings" tab:
1. Hide a variable.  Check all variables still appear on the page.
2. Hide another variable. The variable you hid in step 1 has now disappeared.
2. Hide another variable. The variable you hid in step 2 has now also disappeared.

barrysspace:


danmarsden:
hehe - thanks!

Issue #115, opened 2015-07-12
Send users back to sessions page after creating sessions


danmarsden:
Moved from: https://tracker.moodle.org/browse/CONTRIB-4694

After adding sessions, the user should go back to the sessions page so they know that the sessions have been created. Right now, there is no indication that the sessions were created. An alternative would be the display of some message that the sessions were successfully created.

danmarsden:
copied comment from tracker:
I don't like the way confirmation pages work within attendance - it should reload the same page but display a notification at the top of the page to say something happened... there's a bit of work involved here though, I don't expect I'll have time to work on this as a volunteer but I'm always happy to review patches (or receive funding to work on certain tasks!) - thanks for adding this bug to the tracker.

danmarsden:
this was improved a while back - closing.

PR #171, opened 2015-10-07
Redirection to sessions list after add

antonio-c-mariani:
Hi, Dan

Exemplifies the Moodle mechanism of "notifications" to display messages.


danmarsden:
hmmm - that fixes the UI for users I guess, my concern is that this doesn't appear to be similar to any of the core Moodle code - I prefer to develop code that "might" meet moodle coding guidelines as if a Moodle integrator was reviewing the code and I'm not sure if we'd be allowed to use the $SESSION  var like this. I'll think on it some more and do some digging to see if I can find any Moodle coding rules around the use of the global $SESSION var. I might decide to ignore those issues due to the usability improvements this would bring.

davosmith:
It may be helpful to look at how Totara handles this - the function totara_set_notification(), calls totara_queue_append() ( https://github.com/totara/seedlings/blob/totara-seedlings-2.7/totara/core/totara.php#L593 ) which stores the notification queue directly in the $SESSION variable (a technique I've also replicated inside other plugins I've developed). Later, totara_get_notifications() is called to retrieve an array of messages to display at the top of the page.

Alternatively, you could use the MUC to create a 'session' cache and store the messages in there (which, in reality, will end up storing them in the $SESSION, but just a bit more abstracted away).

danmarsden:
thanks Davo - I'd seen it was similar to Totara, my main issue is that the attendance code is not really structured in the way I would prefer - and if we restructured it properly we may not need to rely on the session var like this. 

I think the likelihood of me finding the time to restructure it is low and the UI improvements this brings are quite good so I'm ok to just go ahead with using $SESSION like this.

Antonio - the class name you've added doesn't follow the frankenstyle naming and new classes should really go into the classes directory. I'm not overly convinced we need a class to do this and could just rely on the simple functions from Totara like Davo points to.  we'd also need to think carefully about the variable name being used "att_messages" isn't quite right  - we should probably use frankenstyle "mod_attendance" but add an extra suffix like "mod_attendance_notifyqueue" so we can avoid conflict with anything that gets added in core.

If you could make those changes it would be great - thanks again!

antonio-c-mariani:
It's done. It also convers the issue #115. Please let me know if you have any other suggestions.

danmarsden:
Looks really great Antonio - thanks heaps for working on this, it's a great interface improvement!

danmarsden:
whoops - looks like behat tests will fail - particularly these lines:
https://github.com/danmarsden/moodle-mod_attendance/blob/master/tests/behat/attendance_mod.feature#L43

let me know if you'll have a chance to run the behat tests and fix them up too.

danmarsden:
FYI - I've added automated Travis CI checks to the repo - see here for the report (which shows the behat tests failing after this patch)
https://travis-ci.org/danmarsden/moodle-mod_attendance

danmarsden:
I think I've fixed the behat failures related to this patch but there are still a few more failing due to other changes - will try to get to them later.

Issue #165, opened 2015-09-28
Prevent the creation of sessions prior the course start date


antonio-c-mariani:
Currently you can create a session dated prior to the course start date. It is automatically classified as "Hidden sessions" not appearing in the sessions list. It sounds good to have an option that controls/prevents this action.

danmarsden:
I'm not sure that I would want to prevent it but it would be nice to show a warning on the add session form that notified the user that the session occurs before the course start date - (using some form of ajax call)

antonio-c-mariani:
Hy, Dan

In our case define a session prior de course start date doesn't make any sense. It means something is wrong: either the course start date or the session date.  In both cases the session simply must not be created. I can't imagine any case where it does. Have you any example where this could happen?

At our university I have changed the plugin code simply including a test at the form validation function showing an error when this happen.  But if it is necessary to keep the current behavior it seems easy define a module setting that control the choice of these options.

danmarsden:
I understand that it shows a potential issue but I can imagine cases where a course starts at a certain time/date but a pre-course meeting might be scheduled were it is useful to track attendance. In any case - it suggests that a problem might exist and it makes sense to warn the user so they can fix the problem but preventing them from creating the session just means they have to go and fix the course start date and then coming back to create the attendance session. It makes more sense in my mind to let the teacher create the session rather than force them to do it all over again.

I'm also not a big fan of adding "settings" to work around behaviour. Sure - it's easy to do at the code level but it doesn't always provide the best UI. One of the common complaints about Moodle is there are too many settings - this is often because the developer has gone through the "easy" route rather than focusing on providing a clean UI.

antonio-c-mariani:
Yes, I agree with you about "settings". They should only be used when they are really necessary. It seems this is not the case.

I'll think a way to warn the user.

There is another related issue. Today a user is able to generate a huge number of sessions just selecting years far from each other when creating multiple sessions. I think it is the case of restricting the difference between sessionenddate and sessionstartdate to 1 (one).

danmarsden:
thanks for the patches - I'm still catching up on stuff after the school holidays - hoping to look at this properly next week!

danmarsden:
merged as part of Antonio's work - thanks!

PR #172, opened 2015-10-20
End date must not exceed one year ahead of the session date

antonio-c-mariani:
See #165.


danmarsden:
Thanks! - I've merged this with a minor change to the language string.

PR #173, opened 2015-10-20
Alert if session date is prior to the course start date

antonio-c-mariani:
see #165.

### DIFF
--- a/add_form.php
+++ b/add_form.php
@@ -48,6 +48,9 @@ class mod_attendance_add_form extends moodleform {
 
         $mform->addElement('header', 'general', get_string('addsession', 'attendance'));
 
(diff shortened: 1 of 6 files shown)


danmarsden:
Looks like a non-invasive way to do it - thanks, I see the error being triggered when I set a date prior to the course start date but if I put a normal date in it seems to completely prevent the page from being posted. 

we can probably remove the console.log statement too - I was using Chrome - let me know when you have a new version to look at - thanks!

antonio-c-mariani:
Yes, it's not working properly. Actually I have liked less and less to use ajax and javascript for this validation. I would like to try a simpler approach. I'll submit a new pull request.

Issue #51, opened 2014-09-25
Scales not working?


t6nis:
Hey,

It appears that the scale option is not working?? Is it meant to be like this or this can be improved?

Use case: When scale is selected then in mdl_attendance the grade field is marked with "-[scaleid]".
Now when marks are added to users then they are not saved in gradebook. Because in locallib around line 890 is:
        if ($this->grade != 0) {
            $this->update_users_grade(array_keys($sesslog));
        }

If the attendance module is not intended to support scales then the scale option in mod_form should be removed to avoid confusion.

SWiT:
I think the scale option should be removed until scales are implemented properly.  I'll make an attempt at a patch at some point soon.

cedict:
We just figured this out as students were getting 'Failed' / 0 in the gradebook. I believe the Attendance module was hidden from student view and gradebook was the reporting method. We had 35 Attendance modules in the system with a scale selected

danmarsden:
Thanks Greg - flagging this as finished, would be good if any of the other reporters on here could verify with the patch I've just merged from Greg.

danmarsden:
Greg's patch caused a regression so I've reverted it and re-opened this. Greg let me know if you'll have a chance to take another look at this.

PR #174, opened 2015-11-02
Added support for scales

greg-or:
Hoping to solve issue #51

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -1234,13 +1234,23 @@ class attendance {
     }
 
     public function update_users_grade($userids) {
+        global $DB;
         $grades = array();
+
+        if ($this->grade < 0) {
+            $dbparams = array('id' => -($this->grade));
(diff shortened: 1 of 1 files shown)


danmarsden:
thanks - looks like it might be doing the right stuff - I've been super-busy so haven't had time to spend on attendance recently - I'll try to find the time within the next couple of weeks!

cedict:
No worries Dan. I just updated the pull request to take into account the att_update_all_users_grades function.

danmarsden:
thanks - looks like it's doing the right stuff so I've merged it across (I presume you guys have tested it at flinders?) :-)

danmarsden:
whoops - looks like there's a regression here:
Fatal error: Using $this when not in object context in /home/dan/www/moodle-r2/mod/attendance/locallib.php on line 1722

PR #177, opened 2015-11-07
Fix issue #165 - Alert if session date is prior to the course start date

antonio-c-mariani:
Please take a look at this approach to address the issue #165. It is much simpler than that using JavaScript.


danmarsden:
even better - great work!

PR #194, opened 2016-04-14
implementing some webservices

caiodoneda:
Hello,

I'm implementing an app to communicate with the Moodle attendance module, and to do so I had to create some web services. Basically, I implemented 3 webservices: the first one retrieves all the today sessions of a teacher; the second one, retrieves the data of a specific session; and the last one, sets a status to a specific user.

What do you think about these changes? What could be improved? I already have some teachers in my university interested in this functionality.

Thank you, and all the best!

Caio Doneda

### DIFF
--- a/classes/attendance_webservices_handler.php
+++ b/classes/attendance_webservices_handler.php
@@ -1,0 +1,142 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


danmarsden:
Hi Caio, 
thanks for sharing your work so far! - adding web-services to the core plugin would be really great!

If you take a look at the Travis CI check shown here in the pull request you will see that your new files do not comply with Moodle coding guidelines - all errors in your files should be fixed and violations/warnings fixed where it makes sense.

I see that you have implemented a new set of functions in the attendance_handler class to return attendance related information - I haven't looked to see if there are existing core functions in the attendance code that will return the same information but because you are implementing new functions that are only called by the webservices API we should add behat tests for the functions in that class before we could merge them with the core plugin.

Feel free to file another pull request if you have been able to address those issues! and thanks again for sharing your work!

danmarsden:
interesting - looks like the travis-ci warning becomes slightly harder to find when you close a pull request! - if you click on the red cross beside the commit id above it will take you to the report for your pull request but here's a direct link to the results as well: https://travis-ci.org/danmarsden/moodle-mod_attendance/builds/123150802

danmarsden:
I'd also suggest you install:
https://moodle.org/plugins/local_codechecker

This allows you to run the moodle guideline checks on your local install - If you are using eclipise/netbeans/phpstorm (and some other IDES) you can configure your IDE to load the coding guidelines and show issues in-line while you are writing your code.

caiodoneda:
Thank you for the suggestions! I'll try to fix the coding style issue and add the behat tests as soon as possible!

Thank you again, 

Caio Doneda

Issue #188, opened 2016-02-08
Totals accumulating over all sessions


peadargrant:
When using Groups and viewing the attendance report, the total for each status seems to be carried over from each session to the next. I am using Groups.

![screen shot 2016-02-08 at 08 59 25](https://cloud.githubusercontent.com/assets/1187400/12881512/443d1ce2-ce42-11e5-8c7c-3451eb9b30d9.png)

Issue #163, opened 2015-09-28
Separate the concepts of "points" and "grades"


antonio-c-mariani:
In our Institution teachers use the attendance module with two purposes: 1) just register and count attendances; 2) grading students based on the attendances. These two forms of use are possible through "grade type" option at attendance module settings. But when the option "none" for "grade type" is selected, the column "grades" is no longer displayed in the reports and teachers can not easily view some kind of summary of the taken attendances.

In this sense, I think it is interesting to separate the concepts of "points" and "grades". Considering this distinction, "Grades" only exist at grade book and they only exist if some kind of "grade type" is selected at the module settings. On the other hand, "Points" always exist as a result of the taken attendances and are always shown in the reports as "points".

danmarsden:
I think the issue that teachers can not see a summary of the taken attendances would be a problem but I don't know that we need to differentiate between grades/points - we just need to make sure the reports show useful information when grades are not being used.

antonio-c-mariani:
Hi, Dan

At #164 you said "We would need to make sure the interface made sense ...". I think this is very import. To be usefull the information needs to make sense. My suggestion goes exactly in this direction. if one teacher choose not to use grades, then grades no longer make sense. Why then keep showing him this term?

When teachers are taking attendances they are not giving grades. Grades are consequence if the teacher choose to use them. In my case (as a teacher) when I see the attendance reports I don't want to see grades. I wonder if students are coming to school or not. To me percentage is the best option (see #38). We have this at our university and our teachers are very happy.

My suggestion is very simple to implement. It can be done just by changing the terms. There are few effective code changes.

danmarsden:
thanks - there are a few open issues related to reports and it would be great to tidy them up a bit more. Screenshots of planned changes can help a lot too.

antonio-c-mariani:
Follows two screenshots of our Moodle production system. This report tries to cover the following issues: #163, #127, #38, #66, #103, 
I'm using two statusset (P4=>4, P2=>2, Au=>0) and (P2=>2, Au=>0). In the session details you can see the acronym and the corresponding points. On the right there are two columns, one for sum of points and another for percentage. At the bottom there is a summary that shows the number of students for each status of the corresponding statusset.
The "Summary" tab is exactly like "All" tab except that it doesn't show session details (columns).
[...]

Issue #127, opened 2015-07-19
Report seems to show incorrect grades when using 2 status set


antonio-c-mariani:
When using two different status set the report seems to show incorrect grades. The attached images shows  a cenario where I expected:
student P 1 - 4.0 / 6.0 - 66.7%
student U 1 - 6.0 / 6.0 - 100 %
but the report showed:
student P 1 - 4.0 / 4.0 - 100 %
student U 1 - 6.0 / 4.0 - 100 %

![1-att_sets](https://cloud.githubusercontent.com/assets/10731910/8767917/89d80694-2e43-11e5-9c96-6df043df1bc5.png)
![2-att_set_1](https://cloud.githubusercontent.com/assets/10731910/8767916/89802cc6-2e43-11e5-80e2-ba1ddff45969.png)
![3-att_set_2](https://cloud.githubusercontent.com/assets/10731910/8767912/8955bcf2-2e43-11e5-9141-6f94ce2efc45.png)
![4-att_ses1](https://cloud.githubusercontent.com/assets/10731910/8767911/8954c86a-2e43-11e5-867f-a9fca3c5508b.png)
![5-att_ses2](https://cloud.githubusercontent.com/assets/10731910/8767915/897ff300-2e43-11e5-8b12-fbe13eca0015.png)
![6-att_rep](https://cloud.githubusercontent.com/assets/10731910/8767913/8956a194-2e43-11e5-9f8a-2caf88d6d4f1.png)
![7-att_grade](https://cloud.githubusercontent.com/assets/10731910/8767914/897a8fe6-2e43-11e5-8b7f-e5c43a8ffc46.png)

danmarsden:
Thanks for the report - the way that table is built is over-complicated and the function that obtains the maxgrade doesn't take into account the status set on each item within the report. It doesn't look like this will be a simple fix and is not something I'll have time to work on as a volunteer.

I'm happy to review any community developed patches or if you have funding/budget to pay for my time to fix this let me know.

antonio-c-mariani:
Hi Dan,

There is a pull request that try to fix this.

Antonio Carlos Mariani

danmarsden:
Antonio - is this one completely resolved now or is there still something missing?

antonio-c-mariani:
It seems ok now.

danmarsden:
cool - thanks.

Issue #38, opened 2014-05-29
Feature Request: Display user's attendance in percentage on report page


nadavkav:
This feature request is actually a regression relative to Moodle 1.9 version of this module.
![image](https://cloud.githubusercontent.com/assets/231168/3122154/f91bd7f6-e763-11e3-87f2-d32a542df617.png)

Please see if you can add this column to the user attendance report.

danmarsden:
this has been completed as part of the work Antonio just completed - thanks Antonio!

PR #195, opened 2016-04-14
Fix report problems and change the way grades are updated

antonio-c-mariani:
This pull request covers the issues #188, #163, #127 and #38. I've tried to separate the report fixing issues from the upgrade grades issues but it's no easy because they are very related.

The function "attendance_get_users_points" calculates the users points. These data are used both to build the report summaries and to upgrade grades. All other changes go around this.


danmarsden:
Thanks Antonio, will be good to tidy up some of the way the grade stuff/reporting is done - I'm pretty busy at the moment and this is likely to require a bit of testing/review - I notice the travis checks are failing on coding guideline issues and a behat test now fails so if you could fix those up too it would be great.

also - I see some big changes to sql statements - have you tested these on PostGres or just MySql? - might be nice to add some more behat tests if possible so we can cover the reports/grading changes too.

antonio-c-mariani:
I have fixed some coding guideline issues. There are many other that were there before the changes.

I do not know much about behat tests. It seems the fail was there before the changes. Could you please check this? I'll try to write new behat tests as you ask.

About SQL statements, some of them was removed and just one is new (attendance_get_users_points function). Yes, I have tested it under PostGres.

Let me know if you have any questions or suggestions about the changes.

danmarsden:
Hi Antonio, thanks for fixing up those coding guideline issues with your code -  the travis checks run on each build and they are passing as of last commit in the main branch: 
https://travis-ci.org/danmarsden/moodle-mod_attendance/builds/120556397

the existing violations and warnings in existing code can be ignored and do not result in a build failure but the behat failures appear to relate to the changes you have made - it's possible that the behat tests are now incorrect and are looking for the wrong data but we cannot merge code that causes the travis checks to report a fail. Ideally your code should not introduce any "new" warnings if possible too. (there are still a few being reported in the build linked above.) 

I'm away most of this week but will try to pick this up again and look at it properly when I get back.

thanks!

antonio-c-mariani:
Hi, Dan

I believe I have identified the failure. I need your help to see what to do.

See the pictures below for the student 3. The student has grade 0/0 because there is no taken attendance for him. In the current version it appears in the "Low grade" report.

gr1

gr2

The condition for the student to be shown at the "low grade" report is set in line 510 of file renderables.php:
   if ($att->pageparams->view != ATT_VIEW_NOTPRESENT || $grade < $maxgrade) {

The value of the variable $maxgrade is set a few lines before:
   $maxgrade = attendance_get_user_max_grade(count($this->sessions), $this->statuses);
that calculates the value 4 for the present scenario. As 0 < 4, the student 3 is included in the "low grade" report.

In the new version, the function "attendance_get_user_max_grade" was removed because it fails when working with multiple status set. The "maxgrade" (maxpoints) value is now calculated by the "attendance_get_users_points" function, resulting in the value 0 (zero) because there are no taken sessions for this student. This seems to be consistent with the value that is displayed in the attendance reports. But because of this the student is not included in "low grade" report, causing the failure at the behat test.

In the new code the line 510 was changed to:
   if ($att->pageparams->view != ATT_VIEW_NOTPRESENT || $points < $maxpoints) {
A simple solution seems to be changing this line to:
   if ($att->pageparams->view != ATT_VIEW_NOTPRESENT || $points < $maxpoints || $maxpoints == 0) {
What do you think about? Do you have any other suggestions?

danmarsden:
thanks Antonio - as mentioned in the other pull request, can you please squash these commits and include the github issue # in the commit messages to tidy it up a bit? - thanks!

Issue #196, opened 2016-04-21
Criteria for "Low grade" report


antonio-c-mariani:
The criteria for a student to be shown in the "Low grade" report is defined by "$grade < $maxgrade" in renderables.php file. In other words this means show all the students who have less than 100% of grade/points.

How about change this criteria and start using "Grade to pass" attendance setting as the threshold for the "Low grade" report?

danmarsden:
makes sense - we've only used the standard grading elements in the form in the master branch (2.8 branch still uses hard-coded grade field) - just didn't update the low grade report based on the new flexibility. Please make sure you submit this as a separate commit rather than including in the bigger one to make it easier to review. Thanks!

antonio-c-mariani:
Hi Dan

I would like to suggest some other related changes (see the picture below):

1) put 'lowgrade' tab at the right of the other;

2) change the 'lowgrade' tab title to "Below n%" (the value of the "Grade to pass" is converted into percentage);

3) add a help icon to the 'lowgrade' tab (only when it is selected)

4) remove the piece of code below from file renderables.php
      
> $currenttime = time();
> if ($att->pageparams->view == ATT_VIEW_NOTPRESENT) {
>           $att->pageparams->enddate = $currenttime;
>  }

   in order to calculate the low grade over all taken sessions.

The last change aims to maintain consistency with the data presented on the "All" tab.

![below](https://cloud.githubusercontent.com/assets/10731910/14746406/decc7758-0885-11e6-8c18-14dbc7e4b466.png)

danmarsden:
I just dug out those changes from your old pull request, tidied them up to squash into a single commit and removed the other changes that were unrelated and put this into the master and 3.5 stable branches here - thanks!

PR #199, opened 2016-04-25
Change lowgrade criteria

antonio-c-mariani:
This pull request covers the issue #196.

I'm sending a new version because there was an error at the behat test in the previous one that I was unable to fix.


danmarsden:
Thanks Antonio - I don't really want to merge the messy commit history there - the commits for some adjusts/other corrections appear to revert changes made between the 2 commits too.

each individual change needs it's own single commit - the lowgrade report changes should be one commit and the other report calculation/grade reporting should be another.

Can you please squash these into single commits and include the relevant github issue # in the commit message as well? - closing this pull request for now as it really needs to be tidied up before it can be integrated. I've taken a look at the changes related to the lowgrade report and I think I'm ok with those, but I still need to review the other report/grading changes further.

thanks,

PR #200, opened 2016-05-03
Webservices

caiodoneda:
Hello,

I've fixed the coding style issues in this new version and I've also added some unit tests to my methods.

Since my functions will only be called via webservices, I suppose there's no reason to create behat tests. What do you think?

All other functionalities remain the same.

Thank you, and all the best!

### DIFF
--- a/classes/attendance_webservices_handler.php
+++ b/classes/attendance_webservices_handler.php
@@ -28,26 +28,27 @@ require_once(dirname(__FILE__).'/structure.php');
 require_once(dirname(__FILE__).'/../../../lib/sessionlib.php');
 
 class attendance_handler {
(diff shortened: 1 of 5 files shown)


danmarsden:
Thanks for sharing the code so far - looks like you have some custom code in there specific to your environmnet related to user profile fields? - can you please remove that and merge your commits into a single commit? - it would also be good if you could create a new issue within the attendance tracker here that describes the webservices you have added and their purpose - then reference the issue id in the single commit message? Feel free to submit a new pull request when you have made those changes. - hopefully that makes sense but let me know if you have any trouble and I can take another look.

danmarsden:
also please make sure that travis does not report any failures with your pull request - and does not introduce any new warning/violations that can be easily fixed (there are some existing violations and warnings in existing code that can be ignored)

PR #201, opened 2016-05-05
Addresses the #188, #127, #163, #38 and #66 issues

antonio-c-mariani:
The previous pull request was really a mess. I was going to remove it and replace it with a more organized version, but you did it before. Thanks.

I have moved the new methods that calculate points from locallib.php to classes/summary.php. This class is used both to update grades and to show summaries in the teacher and student reports.

There are some behat tests for the reports. I have tested them over mysql and pgsql.


danmarsden:
thanks - looking better, really great to see the behat tests too. Have you looked at the performance differences between your code and the old code - particulalry when displaying large numbers of users/attendance data/exporting? - the existing attendance code isn't very efficient on memory or db calls  - particularly inside for-loops when it makes a lot of db calls. I'm still pretty busy and this is a relatively big patch to go through so it might take a week or so for me to complete a review.

danmarsden:
just took a quick look at your patch vs old patch with one of my test attendance sessions - old code 176 DB queries - your code 102 queries - great!

danmarsden:
I'm liking a lot of what I see - thanks for working on this.

what's your reasoning behind no longer using CFG->grade_decimalpoints and the wrapper function attendance_format_float which really just calls core format_float? - I'm not sure that it is correct?

danmarsden:
I ended up having more time to look at this than I expected today - I think once we resolve the issue around attendance_format_float I think I'm generally happy for this to be integrated - thanks again!

antonio-c-mariani:
Personally I do not like the idea of getting spreading parameters among classes, similar to what happened with $decimalpoints, specially because in this code it is a kind of constant. I prefer to concentrate its use in a single point in order to facilitate future changes.

For instance, now we can easily change the function "attendance_format_float" to consider the CFG->grade_decimalpoints global setting instead of using the constant 1 (one):

function attendance_format_float($float, $stripzeros=true) {
    global $CFG;

    return format_float($float, CFG->grade_decimalpoints, true, $stripzeros);
}

On the other hand, the course grade setting "Overall decimal points" no longer seems important due to the focus shift from grades to points in the reports.

How about changing the attendance_format_float function as mentioned above?

danmarsden:
I'd be quite happy to get rid of the use of decimalpoints but if we do that the wrapper function is not needed anymore: - eg:
attendance_format_float($usersummary->takensessionspoints)
should really just be:
format_float($usersummary->takensessionspoints)

I don't think there's much point in adding a wrapper function that really just hides the fact we are just calling the core function - it makes the code harder to read/understand.

antonio-c-mariani:
Ok, this is not a big issue. I've removed the wrapper and replaced it by format_float core function.

There are two other small changes, one related to the query (simplification) in compute_users_points function and the second that adds a new feature to the class 'summary' (very useful in an external context).

danmarsden:
great! - thanks for working on this and sharing the code! - I've merged that in now, I'm planning to release a new version in the plugins db within the next week for 3.1 so it's really great to have this included in that!

PR #204, opened 2016-05-09
implementing web services support.

caiodoneda:
Hello,

I'm opening this new pull request because I've implemented the modifications that you requested.

I've added a new option to indicated where to find the rfid tag, on the settings file, making that information optional now.
I've also fixed the travis failures and squashed the commits to a single one.


danmarsden:
thanks - looking better, but the rfid field is still custom code specific to your installation.
To integrate this into the main attendance code-base you will need to remove it.
An alternative option would be to make the webservice return all user information including custom user-profile fields. Take a look at the other core webservices functions for how they do this - I believe some use the core function user_get_user_details() to populate the data passed back.

PR #205, opened 2016-05-10
implementing web services support.

caiodoneda:
Hello,

I've removed that specific field from the code, in order to make it more generic. I'll leave it that way until I figure out a better way to integrate it to the code.

thank you, all the best.

### DIFF
--- a/classes/attendance_webservices_handler.php
+++ b/classes/attendance_webservices_handler.php
@@ -1,0 +1,130 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+/**
+ *
+ * @package    local_attendance
+ * @copyright  2015 Caio Bressan Doneda
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require_once(dirname(__FILE__).'/../../../config.php');
+require_once(dirname(__FILE__).'/../locallib.php');
+require_once(dirname(__FILE__).'/structure.php');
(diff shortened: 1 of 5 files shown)


danmarsden:
thanks - that's looking better, although I just noticed get_courses_with_today_sessions() looks really inefficient. the calls to get_records("course_modules" and get_record("attendance") shouldn't really be inside those foreach loops - it's better to limit the number of db calls by getting all the relevant records in one query (eg all the course_modules and attendance information for the enrolled courses for the user in one query if possible) - if you could improve that a bit it would be really great. it doesn't have to be one query but the queries shouldn't sit inside the foreach loops.

caiodoneda:
Hello,

This method has been implemented this way aiming maximum use of the Moodle and Attendance module API's. I agree with you that it is a little inefficient and it is using database calls inside the loops, but, due to the way that "mod_attendance_structure" class is instantiated, this was my first approach. 

I could try to reduce those queries to one, and then split the data on different objects, but I believe it would complicate the code understanding. Don't you think?

Although, I was able to reduce the code and amount of database calls. (I'll make the pull request later today)

Thinking about the amount of data that is processed, I believe that further optimization is not necessary. Usually the number of courses are not very big, and we also eliminate the course where the user doesn't have the "takeattendances" capability. After this initial preprocessing, a very small data set will be executed in the second foreach, because usually each course has only one instance of attendance module.

The factor that complicates the optimization is that there may be multiple instances of attendance module in the same course, although this is not the most likely scenario.

However, I have a limited knowledge about the use of this module, since I only know the use at my university. So, if you think that is indispensable the integration of these queries, I won't have problems implementing it.

thank you, and all the best!

danmarsden:
Thanks - some sites have users enrolled in hundreds or thousands of courses and it's never a good thing to put sql queries inside foreach loops that could potentially be called thousands of times.

the calls to get the attendance record and the coursemodule records for each course should probably be easy to replace by a single call to the core function get_all_instances_in_courses() directly after the call to enrol_get_users_courses so that it is called outside the loop and only a single db call is used to get the coursemodule and attendance records for all courses.

caiodoneda:
Hello,

Makes sense. I've implemented this new version, can you take a look?

Thank you, all the best.

danmarsden:
yeah - that looks a lot better! - sorry just noticed "ufscchamadas" here:
https://github.com/danmarsden/moodle-mod_attendance/pull/205/commits/e5741b1c7b344307db47fc42904d7a4e02eb038e#diff-977293f8bda7ae8a9797588f10744016R55

I think that's the last thing that needs fixing - thanks!

caiodoneda:
Hello,

I've fixed that last issue.

Thank you for your time.

all the best!

danmarsden:
great! - thanks

PR #206, opened 2016-06-20
Adding auto creation of calendar events based on sessions

Disss:
Hello.

This pull request covers calendar events making it possible to create calendar events automatically upon sessions adding. Changing sessions' duration, deleting them will affect calendar events.

### DIFF
--- a/classes/structure.php
+++ b/classes/structure.php
@@ -22,6 +22,8 @@
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+require_once(dirname(__FILE__).'/../../../calendar/lib.php');
+
 /**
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks for sharing your code! - please note that all code must pass the CI checks which unfortunately have failed due to some syntax issues and as you are adding a new db field the Behat tests will also need to be updated to allow the checks to pass.

Also - I have a feeling this should be an optional setting rather than creating an event for all sessions. I expect many people using Attendance will not want to put all their sessions into the calendar (but I could be wrong) - you might want to create a post about it in the moodle.org attendance forum to see what other people think? - Closing this pull request as it stands because it causes the Travis checks to fail (see the link here in this pull request for details) thanks again!

PR #210, opened 2016-07-05
Adding calendar events creation, updating and deletion

Disss:
Hi, 
I'm sending a new version of calendar events related functionality. Added/fixed the following:

Settings in attendance activity, determining if moodle should create events automatically on sessions creation or not;
Possibility to create and delete events on sessions page (change duration and delete sessions also works on events);
Added behat tests;
Fixed travis errors.

There's also an issue with icon.svg which gets updated due to line endings edit. I hope it's not a big deal.


brendanheywood:
Cool - I was actually thinking of implementing the calendar stuff myself, I'm curious through what use cases there is for not making the events at all? ie why make this an option and not just always create them. I don't know of any other activity which gives you this option, @danmarsden thoughts?

Disss:
@danmarsden 
Would you mind checking that travis build log? I did not touch any code corresponding statuses and I've no idea what's wrong here.

@brendanheywood 
Hello, I actually did the case you mentioned in the first iteration. But still there can many different use cases and it's always good to have the choice. 
Anyway if a user doesn't want to break the existing ecosystem, he can create events manually (for selected sessions) through sessions page at first, and then with new instances of attendance mod enable auto-creation.

danmarsden:
@brendanheywood - that's probably my fault... I was thinking there would be a case for some sessions to be created that weren't part of the calendar - particularly as we don't put items in the calendar now. But... I can't think of any other activities that give you the choice on this either... I'll think about that a bit more.

travis failure looks weird - strange that it only fails on pgsql too, I'll look into it.

danmarsden:
I've just posted a question regarding this in the forums here: https://moodle.org/mod/forum/discuss.php?d=336015

danmarsden:
it would be good if you could remove that icon.svg change in your patch - if we need to change that file we should really do it with a different commit and details on the reasons for the change.

Disss:
I checked out icon.svg from previous commit. Hope that will do.

danmarsden:
cool - thanks - looks like the travis failure wasn't valid - I reset the individual failure and it gave a successful response on the 2nd run.

Disss:
@danmarsden by the way I saw another failure on the last travis check on pgsql there (different branch and php version though). Do you think it's specific travis related behavior or is there some sort of a bug?

danmarsden:
yeah - I saw that too but refreshing the test seemed to fix it... I haven't seen this intermittent failure before on the attendance code so it will be interesting to see if it occurs again once we merge this into the main branch.

Disss:
I'm closing this pull request as I just understand I forgot to include upgrade.php changes here. I'll add it asap, merge the commits (inc the one in next paragraph) and will make another request.

As for the calendar events use-cases discussion: as long as users don't seem to take part in it (quite odd though), I suggest the following. I will write a script that will create events for upcoming sessions. This script will only work if the setting is on (this setting will be on plugin install/upgrade page and won't be seen too much). Thus users who want events to be created - will have them. As soon as you decide if the events should be created automatically you can just delete the setting on plugin instance page (I still don't like this idea) or just set its default to "Yes". I pretty much think this scenario will do for most of the users. Any suggestions?

danmarsden:
I notice that the choice module in Moodle just ended up with a patch for this (MDL-11369) and they just updated all existing activities to set the events. I think we should probably just do the same - and if people end up with duplicate events from creating events manually in the calendar they will just need to remove them manually. 

So - no setting required, just an upgrade script to update existing sessions.

Disss:
Alright, I think I will do that by the end of the week.

danmarsden:
awesome! - thanks heaps for working on it! :-)

Issue #207, opened 2016-06-26
Users sort order


antonio-c-mariani:
The default sort order when listing users is given by ATT_SORT_LASTNAME. But our preference is ATT_SORT_FIRSTNAME. Currently the teachers have to manually change the list order every time they will take attendances.

This behaviour could be changed by adding a new global attendance config parameter that allows the Moodle admin select this behavior. A different approach is using the "users_order_by_sql" Moodle function as the default behavior.

I prefer the second option because it transfers the sort order to a global Moodle behavior.  I our case we have already changed the "users_order_by_sql" code  to accomodate our needs.

danmarsden:
users_order_by_sql makes sense to me too. - feel free to submit a pull request with that change - thanks!

danmarsden:
thanks - good to have the ordering working better - next step will be to rewrite those tables so they use more normal functions for sorting columns etc.

PR #211, opened 2016-07-07
Changing default users sort order to users_order_by_sql moodle function

antonio-c-mariani:
See #207.


danmarsden:
cool - great to see this, it's something I've wanted to do for a while too - I haven't looked at the code yet but I see a behat failure there - any chance you could take a look at that before I merge it?

antonio-c-mariani:
I need your help because I really don't know what is happening.

The changes are very simple and don't affect (should not) the interface. When I saw the behat failure I run a test locally and it was there. But now I'm not able to reproduce it. Look the picture below that shows the result of the test:

behat_firefox

And here is a print screen where you can see de "Submit attendance" string.

att1

Any idea?

antonio-c-mariani:
I just closed and reopened the pull request and the failure disappeared. It seems there is some problem with the behat tests related to the time when a pull request is submitted.

danmarsden:
looks ok to me - thanks Antonio! - not sure what's going on with behat at the moment.

PR #222, opened 2016-09-09
New functionality: Attendance import

davidherney:
Hi danmarsden...

I developed a new functionality for the attendance module. The functionality add a new tab "Attendance  import", the user can use it for upload multiple users in a specific attendance using a user field (id, username, idnumber, email).

If you decide include it, I can integrate it in the other versions.

Thanks

Saludos


danmarsden:
Hi David - thanks for working on this - I know a lot of people are interested in this feature! - I'm pretty busy at the moment but will try to do a proper review of this in the next few weeks.

in the meantime - could you please squash those commits into a single commit as well? - also it will need to be merged into the "master" branch rather than the 2.9 stable branch as all feature work goes in the latest release (not the older ones) - you might also notice that in the master branch all calls to update_module_button have been removed as it has been deprecated.

thanks again!

danmarsden:
a quick look through the code looks mostly ok - I think I'd prefer that we didn't duplicate the stuff in your new take_student() function - we should really have one central function that does that and re-use it where useful. Also this needs to be rebased onto the master branch and squashed into a single commit with the update_module_button calls removed. - closing this pull request but feel free to submit it again when you have made those changes. Also if you do not think you will be able to make those changes yourself please let us know as someone else may want to pick up your patch and modify it to work correctly.

thanks again for working on it! - a lot of people would find this useful!

davidherney:
Hi danmarsden...

of course, I am going to implement the changes and resend the code in the main branch

Thanks

PR #226, opened 2016-11-08
Attendancebyip

lihidor:
Hi,
We added some code that enables course creators to limit the students to record attendance only from certain IP addresses.
In the form there's a checkbox that you can sign if you want to use this restriction.
Once you marked it you can add specific addresses in the field below. commas separated list, range etc. (as in quiz). We implanted the IP addresses of our LAN as a default.
When the restriction is activated and a student tries to record attendance from an IP that isn't in the list, he Will get a message an will be referred back to the course's main page.
We'll be glad to contribute this code if wanted.


danmarsden:
awesome! - thanks for working on this and for sharing it here!

couple of things:
do we really need the "requirelan" field? - could we just provide the subnet field and assume if it has valid data that it should be used?

Also could you please squash your commits into a single commit? - would also be helpful if your commit message was a short description of the feature - eg "add ability to restrict by IP" instead of "update attendance" etc.

danmarsden:
- also keep an eye on the travis-ci checks that run after you submit a pull request here - those failures will need to be fixed before I can merge the code too.

danmarsden:
I'm going to close this one here as it looks like you still need to do a bit more work on it - (rebasing as a single commit and using a single db field (subnet) instead of the extra "requirelan" field.

You can also get the travis checks to run on your own fork for every commit by enabling travis-ci on your repo at this link:
https://travis-ci.org/taumoodle/

feel free to submit a new pull request when you have made those changes - thanks for working on it!

lihidor:
The truth is the travis-ci is driving me crazy...keeps finding blanks and
tabs.
About "requirelan" - it's purpose is that users that don't involved with
computers (as course creators etc...) that knows almost nothing on IPS and
such will be able to restrict their students in one click without having to
be bothered with "computers' stuff" (because, as I said, the IPS of the
entire LAN is implemented in the "subnet" field as default, so they don't
even have to look at it).

Thank you very much for your quick answer.

I'll do the commits squashing tomorrow or on Sunday.

Lihi

danmarsden:
If it improves usability you could still have that setting in the form I guess - but it doesn't need to be an extra field in the db. I'm not convinced we need it in the form though - have you looked at how the other activites do it? - like the quiz activity access rule? - would be nice to be consistent with that.

you can also run some of the travis tests locally using the moodle local_codechecker and local_moodlecheck plugins available from the plugins db. Thanks for working on it!

danmarsden:
I won't be online much over then next 3 days but will check when I get back to the office.

PR #227, opened 2016-11-10
add ability to restrict by IP

lihidor:
Hi Dan,
This is the new request with one commit and without the "requirelan" field. 
One more diffrence - the default for the subnet field is taken from the attendance configuration, so you can implement it in the mdl_config_plugins table in the database. 
Hope you'll like and merge it.
Lihi


danmarsden:
thanks - looks like the travis checks are picking up a few issues with the behat tests - particularly this doesn't look good:

 Moodle exception: Exception - Notice: Undefined property: stdClass::$subnet in [dirroot]/mod/attendance/mod_form.php on line 65 

are you able to take a look at those?

danmarsden:
I just re-ran the travis checks on the master branch and I see something in the last 2 months has changed in the Moodle code-base causing some failures in the existing branches - I'll try to get those fixed in the next day or so.

danmarsden:
I've fixed up some of the existing issues - not sure what's going on with the 5.6 PHP tests - doesn't make a lot of sense to me but Travis is also having issues at the moment and timing out on checks that should pass. - if you could fix the error reported in the PHP 7 tests:
 Moodle exception: Exception - Notice: Undefined property: stdClass::$subnet in [dirroot]/mod/attendance/mod_form.php on line 65 More information about this error
and then resubmit again we can probably ignore the other failures for now.
thanks!

danmarsden:
also looks like you've removed the timemodified field in the install.xml file accidentally, and it might make sense to check if subnet is not empty before passing it to address_in_subnet()

managed to get php 5.6 tests running locally here and no failures reported when running local - so must be something wrong with travis ci checks at the moment.

lihidor:
hi, 
I added back 'timemodified' (sorry for that) and a check if the IP address is empty.
Sorry, but I'm really having troubles working with Travis-ci. Can't understand where ans why it founds errors and hence can't fix them. You are invited to fix them yourself if you please.

danmarsden:
I've had a quick look at this, rebased your latests commits and tidied a few things up here:
https://github.com/danmarsden/moodle-mod_attendance/tree/requiredip

but... while looking at this further I wonder if it should really be an activity-level setting. I can imagine many cases where a course would be split across multiple locations/labs and each session may have it's own need for a subnet range. What are your thoughts on this?

lihidor:
For us, because people can  actually connect from all over the campus, even from mobiles with Wi-Fi, the critic thing is that the address will be inside campus, and not the specific location. this is why, as I told you, we implant all the network as default. 
If it doesn't suit others feel free not to merge it.
Lihi

danmarsden:
It's definitely useful as it stands, and I've merged it into the master branch manually. Thanks for working on it! Support for setting this at the session level could be added in future. By default it would use the setting made at the activity level but should allow it to be overridden to use a different subnet.

feel free to submit any other improvements as pull requests!

lihidor:
Glad you merged it. 
Thank you.

Issue #213, opened 2016-07-25
lock the student names so they float (enhancement)


reactor0138:
That is to lock the student names so they float. On your attendance block, this is on your report tab, where it becomes near impossible to check attendance after you have add 20 or more slots (unless you have 2 screens or a really high res setting on your display). It would work like Gradereport does where the name stay as you scroll across the page.

danmarsden:
I agree this would be nice - feel free to help us work on this improvement!

jstilwell:
Thirded. This would be a very nice improvement to the interface for instructors.

danmarsden:
Jesse - feel free to help us work on it! :-) - it's pretty high on the list of things I'd like to do if I have some free time to work on feature dev for the attendance.. unfortunately my free time usually gets taken up by general maintenance tasks etc.

danmarsden:
I've just pushed a patch through that does this using css in the master and 3.5 branches here in github - seems to work well in my tests, but happy for feedback - thanks!

PR #230, opened 2016-11-29
Taken sessions by acronym

antonio-c-mariani:
Code proposal for # 225.

I don't know what the users really do with this data. Thus, the best choice seems to be including them in all report tabs. This is the simpler way to implement this too.

In the future would be nice if there is a way the user could  select what data he would see (sessions details, sessions by acronym, summary) or just implement #213.


danmarsden:
cool - thanks! - I'll take a look at this now

danmarsden:
pulled this separately, rebased and added a few small changes to keep travis happy. Thanks again for work on this!

PR #233, opened 2016-12-06
Add missing search:activity string.

dannyjung90:
The string for search:activity is missing from the language file.


danmarsden:
thanks for the pull request!

Issue #224, opened 2016-10-20
Attendance Totals in export/move row total columns.


haietza:
We have had requests to
- Move the row totals columns to be next to the student names on the Report tab, All view
- Include the row totals in the export file

danmarsden:
I'm not sure about moving the row total columns, but if that was configurable it might be ok. Feel free to submit a pull request - or contact me if you are able to fund the time for me to do the work. (I work for the Moodle Partner Catalyst IT)

antonio-c-mariani:
Hi

Here are some suggestions related to the first request (see the picture below):
* Divide the header into two rows;
* Add two buttons in the sessions column, the first one to hide/show session details and the second to move the entire session details to right (after the "Taken sessions" column) and left again;
* Move the "Send a message" button to the end of the message sending column.

What do you think about this?

![attendance](https://cloud.githubusercontent.com/assets/10731910/20991468/cc39a35a-bcc5-11e6-979d-720184b9475a.png)

danmarsden:
I'm ok with 2 rows for the header, but I'm not sure about those buttons - seems quite different from other tables in Moodle doesn't it?  often tables like this in Moodle have a "preferences" section underneath the table that allows different display options to be changed which might be better? - I haven't taken a close look at any of the newer tables in 3.2 though so if there are interface ideas we can re-use to be consistent with other forms let me know!

antonio-c-mariani:
The inspiration was the grader report (/grade/report/grader/index.php) in which buttons are used to control the detail level of the grade aggregations. The same approch is used by mod/assign when viewing submissions to hide/show columns. The same icons are used here to hide/show session details in order to keep semantic consistency. 

When an action is related to a specific column (column sorting, hide/show details, etc) I think this approach of icones (or links) associated to the column title is better than the "preferences" section. The second approch seems better for filters and for other actions and preferences that are applyed to the overall table or to a group of columns.

danmarsden:
I've seen the hide/show stuff before (that was fine and makes sense sorry I wasn't clear there.) but I haven't seen the arrows used to change the order of columns before - is that used somewhere else that you know of?

danmarsden:
it does look like a cool idea to re-order the columns that way - I just want to be wary about deviating from the "normal" tables in Moodle.

antonio-c-mariani:
I didn't find any place where the arrow has this semantic maybe because is not common to have this behaviour of changing column positions in tables.

[...]

PR #237, opened 2016-12-15
Report changes

antonio-c-mariani:
Issue  #224


haietza:
@antonio-c-mariani @danmarsden Our users are really pleased with the movable session totals columns - thanks to you both!

PR #238, opened 2016-12-17
Move hide session details

antonio-c-mariani:
Hi, Dan

This pull request try to cover the issue  #224. It includes some other minor fixes.

Please note the help icons in the summary report tab. They try to explain the semantic of the three percentage columns.

Let me know if there is something that must be changed. Please correct the English text as you wish because it's not my main language. 

A.C.Mariani


danmarsden:
I like this a lot - thanks heaps! - typically column hiding would be done with user-preferences instead of for all users but that could come at a later date if people request it. I like the new help icons and will make a few small adjustments to the language - thanks again for submitting such a helpful patch!

PR #244, opened 2017-03-07
Fixes #243 - Moodle 31 - Global Search issue with Database Schema

### LINKED ISSUE #216: latest version breaks global search in 3.1
Error reading from database Debug: ERROR: column "timemodified" does not exist
LINE 1: SELECT * FROM mdl_attendance WHERE timemodified >= $1 ORDER ...
Looks like every module needs to have a timemodified

### LINKED ISSUE #243: DB Schema does not have intro and introformat
Hi Dan,
I've been doing some testing with this plugin for Moodle Global search. The codebase I'm testing on is 3.1 but it looks like it will exist on Moodle 3.2 also.  This is also possibly related to Issue #216 

According to the Moodle dev docs (https://docs.moodle.org/dev/Activity_modules#install.xml) activity plugins need, `timemodified`, `intro` and `introformat` in their DB schema. mod_attendance has had timemodified introduced but doesn't have the other two. This is causing errors when indexing mod_attendance documents in global search. AFAIK global search is currently the only core code that is enforcing this.

Global search gives the option for activities to override the get_document method which would be one way to solve this, however, I'm guessing it should have the req'd fields in the DB.

I cherry picked the commit from Issue #216 into the 3.1 version of the plugin as part of my testing. As an aside will this be back ported into the MOODLE_31_STABLE branch of the plugin?

roperto:
To replicate the issue, please check #243.

The first commit (fixes #216) I just backported it.

* Performed a fresh install of the mod (not using upgrade).

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -1,8 +1,9 @@
+
 <?xml version="1.0" encoding="UTF-8" ?>
-<XMLDB PATH="mod/attendance/db" VERSION="20130704" COMMENT="XMLDB file for Moodle mod/attendance"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:noNamespaceSchemaLocation="../../../lib/xmldb/xmldb.xsd"
->
+<XMLDB PATH="mod/attendance/db" VERSION="20160829" COMMENT="XMLDB file for Moodle mod/attendance"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:noNamespaceSchemaLocation="../../../lib/xmldb/xmldb.xsd"
+        >
   <TABLES>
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks Daniel - version bump doesn't look right for a stable branch - only the last digit is supposed to be incremented in stable - (date should only change on the master branch)  also take a look the travis failures - thanks!

roperto:
Hi Dan.

I am checking the failing tests.

Regarding the version bump, if I only do a minor bump how is the upgrade.php going to properly identify if it needs to run?

Cheers,
Daniel

roperto:
I think I know the answer for my own question: make upgrade.php prepared only for that branch. I will make the PR again soon.

danmarsden:
Cool thanks. All sorts of bad stuff can happen if you change the full date version in stable branches.

PR #246, opened 2017-03-07
Moodle 30 - Global Search issue with Database Schema

roperto:
Moodle 30 has no search plugins, so the global search is not a problem.

As stated in #243 the schema still diverges from the documentation.

* Performed a fresh install of the attendance mod (not using upgrade).

@danmarsden please let me know if I need to change something

Cheers,

Daniel

### DIFF
--- a/classes/structure.php
+++ b/classes/structure.php
@@ -50,6 +50,15 @@ class mod_attendance_structure {
     /** @var float number (10, 5) unsigned, the maximum grade for attendance */
     public $grade;
 
+    /** @var int when was this module last modified */
+    public $timemodified;
+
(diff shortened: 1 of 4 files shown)


danmarsden:
I'm not convinced we need this on the 3.0 branch - typically db level changes are avoided on stable branches - and the dev docs are not usually version specific. If we aren't fixing any actual bugs I'd probably prefer to leave the other branches as-is.

roperto:
Hi Dan.

From the perspective of the search (what we need now) this is not needed at all.

I fixed the typo just but actually I am closing this one. We can reopen if needed later.

Cheers,

Daniel

PR #248, opened 2017-03-07
Moodle 32 - Global Search issue with Database Schema

roperto:
Moodle 32 presented some warnings when indexing for a search plugin.

* Performed a fresh install of the attendance mod (not using upgrade).

@danmarsden please let me know if I need to change something

Cheers,

Daniel


danmarsden:
same issue - small typo in comments referring to timemodified - otherwise looks good.

Issue #262, opened 2017-05-09
Data and Time is not visible after creating session in sessions list


rajeevdubey007:
Hi Currently I am using attendance module version '2016121303'.
Date and Time are not shown in session list page:
![image](https://cloud.githubusercontent.com/assets/13365691/25849629/d78cb0a0-34dc-11e7-8136-832679e96d5a.png)

Is there any setting I need to set? Please help.

danmarsden:
I haven't seen this before - have you made any changes to the language pack for the attendance plugin?

rajeevdubey007:
No, I simply installed the attendance plugin and try to create a session but after creating session I am getting the list without date and time.

danmarsden:
Probably a compatibility issue with the parameters we're passing to strftime (function used to convert unix timestamps to user readable times.) What operating system is your site hosted on? - is it a Mac OS? - I just noticed we're using %P in the strftime call which looks like it is not supported on Mac OS.

rajeevdubey007:
Hi Danmarsden,

We are using "Windows Server 2008 R2 standard". 
Should I change timestamp in code somewhere in order to work on my server?
Or if you have any idea please suggest.
Thanks.

danmarsden:
as much as I'd like to encourage you to use a different O/S I guess we should support those unfortunate souls forced to use Microsoft systems.... :-)

here are the strings that will need changing:
https://github.com/danmarsden/moodle-mod_attendance/blob/master/lang/en/attendance.php#L278
https://github.com/danmarsden/moodle-mod_attendance/blob/master/lang/en/attendance.php#L280
https://github.com/danmarsden/moodle-mod_attendance/blob/master/lang/en/attendance.php#L281

Try changing the capital P in those lines to a small p and then clearing site cache to see if that helps.

if that isn't it, try changing the lower case l(for lima) to a capital I (for India) - and clear site cache.

let me know which combination of those works.

rajeevdubey007:
Hi Danmarsden,

I tried every combination but it didn't work. Even I also installed the same module in local system that uses windows 7 but there I am getting the same issue.

![image](https://cloud.githubusercontent.com/assets/13365691/25948973/7f1dd082-3673-11e7-9ecc-66493836ad17.png)

[...]

PR #272, opened 2017-09-01
Fix behat failures

gthomas2:


danmarsden:
Do you have context for this? Might need to be aware there's currently a failure in master branch due to a regression in formslib related to the new hiddenif stuff.

danmarsden:
the fact that you've changed the case of the AM/PM text in the behat tests makes me wonder if you have made local changes to the lang files that are causing failures for you. (or maybe you're running behat in en_us instead of en?) looks like behat is failing to run on your changes (not sure why) but I suspect if they run they will fail on the core code.

gthomas2:
I just downloaded the plugin and tried running the behat tests for it on a vanilla Moodle 3.3.

For the AM / PM case sensitivity issue -perhaps the 'AM / PM' text is generated by PHP date/time formatting and there could be some variance between either PHP versions or the default locale for PHP. I will work on a better fix for this today.

I also found another issue where Behat would try to enter attendance data before the form was available, so I added this step to work around that:

I wait until attendance table is visible

gthomas2:
I have two completely separate environments with PHP installed, one on osx and the other on ubuntu.
On osx %p results in "am" and on ubuntu it results in "AM". The CI environment where I work resolves %p to "am" but the development image we use resolves %p to "AM". I can write a custom step which will work with either.

gthomas2:
After some digging it seems like PHP support for %p is hit and miss.

http://php.net/manual/en/function.strftime.php

"Not all conversion specifiers may be supported by your C library, in which case they will not be supported by PHP's strftime()"

gthomas2:
I've fixed the behat failures in a way that will accept %p being output differently in some environments. The alternative solution to this would be to modify the existing code to force the case of anything formatted with %p.

danmarsden:
Thanks for the context - I hadn't seen this because Travis runs on ubuntu and we run all our CI tests on ubuntu here at Catalyst too.  I'm not sure what I think about "I wait until attendance table is visible" - (is that really needed and under what conditions can you reproduce that as an issue?)

it sounds like we might need to change the %p to something else if it's not very reliable. We changed them recently in #262 to improve compatibility with Windows machines - but I didn't look at compatability with MacOs.

gthomas2:
I was wondering if we could swap to using the PHP date function to get 'am' or 'pm' but sadly it does not localise.
However, after searching for problems with people using %p it seems that it does not localise for certain languages correctly - e.g. Spanish:

https://stackoverflow.com/questions/6852947/strftime-doesnt-show-pm-or-am

So the localisation benefits of using strftime to print out "AM" or "PM" are pretty limited.

danmarsden:
you can change it in the lang packs for languages that don't support %p - but it sounds like it's not just locales that don't support it but the Web server itself that may not support it either.

gthomas2:
For the "I wait until attendance table is visible" step, it fixed an intermittent failure and I can't reproduce it now. Basically mink would try to click on things in the attendance table and it couldn't because the attendance table wasn't available. I could probably remove the "I wait until" steps if you like.

danmarsden:
yeah - I'm not a fan of adding code "just because" so the "I wait until attendance table is visible" is probably a non-starter. Also - although your patch will make the tests pass on a wider range of systems I'm not sure it's really the right way forward. We might need to modify the strftime string to something that is more cross-platform compatible. Issue #273 is a recent one from a user on CentOS where the %p doesn't appear to be working either.

danmarsden:
here's what I was thinking might be a better method:
https://github.com/danmarsden/moodle-mod_attendance/commit/6fa7ca8655a28bb46386c69d616360ad93a17364

what do you think?

danmarsden:
- with a clearer head, that patch isn't quite right because each locale may use a different value for AM/PM - what we might need to do is check if %p is included in the string, remove it and compare the result with and without the %p - if the result is exactly the same (with and without %p) then add the english version of AM/PM depending on the %p or %P - make sense?

gthomas2:
Yes, that makes sense. I like the approach ;-)

danmarsden:
cool - closing these pull requests and I'll try to fix that patch up today - thanks for the report and discussion!

gthomas2:
No worries! And kudos to you for having behat tests that revealed a problem ;-)

danmarsden:
2nd time this week that the behat tests have picked something up which is nice for a change! - it picked this up in the moodle master branch: https://tracker.moodle.org/browse/MDL-60012

PR #280, opened 2017-11-21
Add shortname to services file

willianmano:
Adding the shortname we can now retrieve an user token via an external application using webservice.

### DIFF
--- a/db/services.php
+++ b/db/services.php
@@ -51,8 +51,15 @@ $functions = array(
 
 
 // We define the services to install as pre-build services. A pre-build service is not editable by administrator.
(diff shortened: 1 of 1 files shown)


danmarsden:
looks ok to me - merged into master, will cherry pick into other branches later. Thanks.

willianmano:
Thats good. :)

I am developing a mobile application to do checking in . Now I am localy changing your plugin to print the name tags with a QR Code with the user id.

danmarsden:
sounds interesting! - feel free to file another pull request if you think the work you are doing might be useful for anyone else! - thanks.

PR #288, opened 2018-01-17
mod/attendance: Replace disabledif with hideif

Alicecj:
I have replaced all instances of disabledif with hideif and have **not tested** it at all on any versions.


danmarsden:
Thanks for the pull request @Alicecj - I did some testing and found that the disabledif rules operate slightly differently from hideif when working on grouped elements so I've merged your commit and added a patch on top to adjust the rules.

PR #296, opened 2018-02-22
Add description to event, and group to event name if present.

nwp90:
It seems to us that there is no good reason not to use session description as calendar event description.

It is also extremely useful to have group name in calendar event name, as there are several possible use cases where only the event title will be displayed (e.g. when calendar exported to various external calendar progs), and without group name in event name all the different groups' events appear to be duplicates.

### DIFF
--- a/classes/calendar_helpers.php
+++ b/classes/calendar_helpers.php
@@ -52,9 +52,14 @@ function attendance_create_calendar_event(&$session) {
     $caleventdata->instance       = $session->attendanceid;
     $caleventdata->timestart      = $session->sessdate;
     $caleventdata->timeduration   = $session->duration;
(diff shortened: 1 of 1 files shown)


danmarsden:
can't think of any reason either - have cherry-picked to supported stable branches and added another commit to make sure descriptionformat is included when creating the event as well. Thanks for the pull request!

danmarsden:
thanks - I don't like the general UI that hides/shows/moves the columns in that report but I don't think I've seen a better core method yet and it does provide useful functionality and the way you've implemented it is consistent with the existing UI.

Maybe one day we'll make it look nicer...

cherry-picked back to Moodle 3.2 and higher. Thanks!

PR #310, opened 2018-06-25
Make events optional per-session.

### LINKED ISSUE #309: Event creation should be optional per-session
We need calendar event creation to be determined by a per-session setting. PR coming.

nwp90:
Fix to allow creation of calendar event to be controlled per-session. Fixes #309.


danmarsden:
thanks Nick - looks interesting - I'll need to get this into the master branch first and can then cherry-pick into older supported releases. I'm quite busy this week but will try to look at it asap!

nwp90:
Works fine as is, at the moment the default is fixed - that could perhaps best be a setting on the plugin.

PR #324, opened 2018-08-22
Issue22 - Add QR code to password, to speed up attendance taking

ecampbell:
This is a set of code changes to support QR codes. It adds 1 field to the attendance_sessions table, includeqrcode, which is a flag to control whether a QR code should be displayed on the password.php age. It also includes a custom copy of attendance.php called attendance_qr.php, which automates the student attendance marking process so that they only have to log in to Moodle, and they are marked as present.
The attendance_qr.php contains a lot of redundant code which I kept in for easier comparison, but it should probably be removed. Alternatively, attendance_qr.php could be merged into attendance.php, but that would make it more complicated.

### DIFF
--- a/password.php
+++ b/password.php
@@ -25,7 +25,9 @@
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
+use Endroid\QrCode\QrCode;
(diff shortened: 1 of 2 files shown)


danmarsden:
closing this as it is superseded by pull #326

Issue #329, opened 2018-09-10
Improve status set editing form


danmarsden:
as mentioned in issue #328 

when creating status sets: when you try to click "add" and only entered "acronym" and "description" for the current item, you will get "error writing database"

but in general the UI in that page needs a lot of work - the "add/update" buttons should really be combined - a call to "update" should really insert any new records entered, and proper validation should be added to the form at some point.

PR #335, opened 2018-10-11
SUP-7832 Fix bug in mod_attendance defaultstatus

T4m:


danmarsden:
thanks for the pull request - I've gone for a slightly different method to fix this which also prevents non-numeric data being added in the form. I really want to rewrite this form at some point (see Issue #329 for some of the other issues.)
thanks!

PR #355, opened 2018-12-18
Fixes report table row height for multi-line user names.

mwehr:
Hi,

This fixes report table row height for multi-line user names.
Switched name part to checkbox label element.

regards,
Mario


mwehr:
Hi Dan,

You are right normally the table row should adjust is height according to the chields height.
The problem here is that the first td element with the user names is absolut positioned. This elements are removed entirely from the document flow. That means they have no effect on their parent element.
Because of this I came up with the JS solution :-)

danmarsden:
sure - but the attendance isn't the only activity that uses that sort of positioning for names - I'm still not convinced we need this sorry - maybe you could create a tracker issue (with screenshots) here that shows the problem you are trying to fix?

mwehr:
Hi Dan,

I just want to clarify that my fix is only for the special handling of the report table element in your attendance report view. Normally, and this also applies to moodle, the JS is not needed to handle element height adoptions according there chields height.
For example if you look into the grading table which has a similar feature. This one is a completely  different implementation which don't uses tabel elements.

danmarsden:
fyi - this got closed "automatically" as I renamed the "master" branch to "main" - I'll see if there's a way I can reopen this soon.

PR #372, opened 2019-02-28
Webservice: allow multiple instances in one course

gormster:


danmarsden:
thanks - looks like a semi-obvious bug there!

PR #376, opened 2019-03-26
Fix #357 - Added QR code icon

### LINKED ISSUE #357: Improvement: Use QR code icon for the QR code feature instead of the key icon
Hi Dan,

when you introduced the QR code functionality in your plugin, I had problems to find the QR code when using the feature.
I was searching for some icon that indicated to me that I could find the QR code there. It took me quite some time to press the key icon in order to find the QR code behind it:
session
 
So my proposal is to use a QR code icon (e.g. https://fontawesome.com/icons/qrcode?style=solid&from=io) instead to make it more obvious for the user what he will find behind the icon.

Best, Kathrin

maksudr:
Added logic to check if the QR Code option has been checked and uses the font awesome QR code svg. Closes #357 

image


danmarsden:
Hi @maksudr, I've just tested this and I see some php notices for sessions that do not have the qrcode enabled, and it doesn't seem to actually replace the key image with the qrcode image either - looks like you haven't tested this specific patch?

When doing development, make sure you have debugging mode set to developer with "display errors" set to on. I tend to do this on my config.php with the following block of code in my config.php

@error_reporting(E_ALL | E_STRICT); 
@ini_set('display_errors', '1'); 
@ini_set('log_errors_max_len', '10000');
$CFG->debug = E_ALL; 
$CFG->debugdisplay = 1;
$CFG->noreplyaddress = 'noreply@danmarsden.com';
$CFG->noemailever = true;
$CFG->showcrondebugging = true;

maksudr:
Hi @danmarsden, Thank you for flagging this. I missed a few lines when I committed the patch. I have also addressed the PHP sessions. This should be fixed now. Cheers!

danmarsden:
Great - thanks @maksudr - merged! :-)

Issue #351, opened 2018-11-25
QR Code password rotate/expiry


danmarsden:
It should be possible for the QR code password to expire/rotate.

eg - each qr code password could be valid for 30seconds, and every 15seconds the qr code would refresh. (values possibly configurable at the site admin level) We'd need to be careful about how to deal with attendances where the user is able to select their own attendance level and make sure the code is validated on the first page load - not on the save of the attendance session form.

We would need to modify the code to store multiple passwords for each session (instead of just a single password as is currently stored) - each password hidden in the QR code would need an expiry time and the page would ideally use ajax to update the qr code image.

danmarsden:
For more context this was also mentioned here:
https://github.com/danmarsden/moodle-mod_attendance/issues/323#issuecomment-420494818

and here:
https://moodle.org/mod/forum/discuss.php?d=378173

danmarsden:
NOTE: if an external pull request is submitted with this we'll need some form of automated tests to cover it. As this relies on timing values I suspect unit tests would be better than Behat but I'm open to either.

MonilBhavsar:
I would like to work on this issue.

danmarsden:
Note to GSOC students - please make sure you have read:
https://docs.moodle.org/dev/GSOC#Applying_to_be_a_GSoC_student_with_Moodle

If you have any questions about the attendance module in general please use the Attendance forum here:
https://moodle.org/mod/forum/view.php?id=741

If you have any GSOC specific questions, the best place to ask them is here:
https://moodle.org/mod/forum/view.php?id=8360

Please do not contact me privately - all communication from potential GSOC students must be done using the public communication channels - this helps to maintain a fair and transparent process for all students.

ScSherifTarek:
Hi Mr. Dan, I'm Sherif Tarek third year student of Computer Science. I'm preparing to join with you in the "Attendance password rotation/expiry" project s a part of GSOC 2019. what I understand that you need to make the validation just in the first step then he can go on the other attendance steps. what I can't understand is what I assigned to do should I implement that or you have already done it?!

danmarsden:
@ScSherifTarek - it sounds like you need to do some more research - you should get a local Moodle installation running with the attendance plugin, read some of the (unfortunately limited) documentation on the attendance activity and see how the existing QR code functionality currently works. This may help you to better understand what is required. Good luck!

abhimanyuZ:
Hello sir,

> to store multiple passwords for each session

[...]

PR #380, opened 2019-03-30
#351 Added Attendance password rotation/expiry

Bellangelo:
It does not save old passwords or expire them because it shows a timer when the password will be changed.


danmarsden:
Thanks - it doesn't look like you have understood the requirements here - if you are trying to improve your GSOC application you should work on some other unrelated tasks that demostrate your ability to understand the issue/problem solve and follow our guidelines/procedures. Thanks.

Bellangelo:
Thank you for your comments. Could you please elaborate what is wrong with my code? Or you mean that this project should be a full time project for the summer?

danmarsden:
Yes - this was supposed to be part of the project work, but the main issue not considered here is server performance and only storing 2 "passwords" - I'd prefer to see multiple passwords generated and then rotated without requiring a call to the server every time the QR code changes - I'm happy to provide more feedback such as coding guideline issues, namespacing issues, how to handle inclusion of 3rd party code, but this also means that you will need to think about what to include in your project proposal if all the work is completed prior to GSOC. 

thanks!

Bellangelo:
Thank you very much for your help! I now understand more the project. Thanks again!

PR #381, opened 2019-03-31
Fixes #378 - Enable QR Code without filling Student Password field

### LINKED ISSUE #378: Cannot enable QR Code without filling Student Password field
A user can leave the "Student password" field empty and tick the "Include QR code" box and still save the session. This behaviour is misleading to the user and gives the impression that the attendance can be taken with the QR code only. The user will not see the option to display the QR code and has to troubleshoot to figure out that they must also fill in the Student Password field.

I believe the page should flag an error and prompt the user to fill in the student password field and not let the user save until the field has been filled.

A video demonstrating the bug:
[Video Link](https://gfycat.com/safespectacularhairstreakbutterfly)

maksudr:
This PR Fixes #378. The user is now able to have a session with no password but still provide the QR code to have students sign in.

**Changes:**
**attendance.php**
- There was a requirement for there to be a 'sesskey' present to view attendace.php page, this has been removed. Viewing the page itself does not change state, so it is unnecessary for there to be a 'sesskey'. The token is important only for the POST request (form submission) which does change state, to prevent CSRF. It was also a blocker to the solution as the user needed a 'sesskey' to get directly to the page. The 'sesskey' is still generated in the form on page load and sent in the form submission to prevent CSRF. There are no security implications to not requiring the 'sesskey' to view the page as the user needs to be authenticated, enrolled in the course and be visiting the page during the sessions attendance time.

**render.php**
- The page now shows the password icon if the session has no password and just the QR code. When #376 is merged, this will be a QR code icon.

**password.php**
- Code refactor

The page now accounts for all 3 session situations:
- Password and QR Code
- Password only
- QR Code only

### DIFF
--- a/password.php
+++ b/password.php
@@ -49,15 +49,16 @@ $PAGE->set_context(context_system::instance());
 $PAGE->set_title(get_string('password', 'attendance'));
 
 echo $OUTPUT->header();
-echo html_writer::tag('h2', get_string('passwordgrp', 'attendance'));
-echo html_writer::span($session->studentpassword, 'student-password');
 
-if (isset($session->includeqrcode) && $session->includeqrcode == 1) {
-    $qrcodeurl = $CFG->wwwroot . '/mod/attendance/attendance.php?qrpass=' . $session->studentpassword . '&sessid=' . $session->id;
-    echo html_writer::tag('h3', get_string('qrcode', 'attendance'));
+$showpassword = (isset($session->studentpassword) && strlen($session->studentpassword) > 0);
+$showqr = (isset($session->includeqrcode) && $session->includeqrcode == 1);
 
-    $barcode = new TCPDF2DBarcode($qrcodeurl, 'QRCODE');
-    $image = $barcode->getBarcodePngData(15, 15);
-    echo html_writer::img('data:image/png;base64,' . base64_encode($image), get_string('qrcode', 'attendance'));
+if ($showpassword) {
+    attendance_renderpassword($session);
 }
+
+if ($showqr) {
+    attendance_renderqrcode($session);
+}
+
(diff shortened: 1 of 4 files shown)


danmarsden:
Thanks @maksudr - sorry for the delay in reviewing this one!

I like the idea of putting the qrcode/password rendering into their own functions - I can see us needing them in future.

But - functions shouldn't be embedded in in-line files and these 2 could probably be moved into the locallib.php file (make sure you use the frankenstyle prefix on the function names too)

Also - $CFG shouldn't be passed as a var - just do a global $CFG inside the function to get access to it.

I would also be inclined to make this block a lot simpler:
https://github.com/danmarsden/moodle-mod_attendance/pull/381/files#diff-01e5c1d8a92dfd2745e7c88c51bcc9a6R89

eg:
if ($showpassword) {
  //print password
}
if ($showqr) {
 // Print password
}

I'd also be tempted to just pass the full "$session" var to the renderqrcode function and then inside that function check if studentpassword is set or not, and use it if needed.

Thanks heaps for taking a look at it! :-)

maksudr:
Thank you for reviewing this @danmarsden. I'll action these changes this week :)

maksudr:
@danmarsden I have actioned the changes :+1:

password.php
```diff
@@ -29,6 +29,7 @@
 require_once(dirname(__FILE__).'/locallib.php');
 require_once($CFG->libdir.'/tcpdf/tcpdf_barcodes_2d.php'); // Used for generating qrcode.
 
+
```
danmarsden: hmm - I would have expected this change would have triggered a travis failure - we're not supposed to have 2 empty lines in the code - when you get a chance can you please remove this extra line - thanks!

Issue #365, opened 2019-02-13
when studentscanmark is disabled, automatic marking is still available but doesn't work.


danmarsden:
when studentscanmark is disabled, the session editing screen allows the teacher to select automatic marking, but the status set config page hides the "available to students" field so automatica marking cannot work.

we need to look at either hiding the automatic marking options, or showing the available to students field even when studentscanmark is off.

gormster:
Alternatively: automatic marking should work even when studentscanmark is off… our use case is entirely webservice based, meaning that you often end up with partially complete attendance sheets. Currently we advise instructors to mark every student as absent before starting a class, but that's not an ideal solution for obvious reasons. Automatic marking seemed like a great answer for us… except that it doesn't work when self-marking is disabled.

I can't really see any reason why this would be the case - the aspect of automatic marking we really need is a default status that's automatically set for every student for whom no response has been marked, and that doesn't depend on self-marking at all.

What are your thoughts on this? If `update_user_status` had an option to specify an "automatic" status - that is, the status that would be applied if the student were marking themselves - then more-or-less all of automatic marking could be achieved through web services, from a device with no inputs (like our card readers).

danmarsden:
I think the issue is mainly with the status set settings ui which simplifies when can mark is disabled. I suspect we just need to always show all options on that ui even if can mark is disabled. There may also he some ui/ux improvements to make the settings more clear.

danmarsden:
We'll also need to improve the session add/update forms so that automatic marking can be set independently of the student marking. I'll see if I can get to that this week.

lszeremeta:
Hi,
I have two suggestions:

**_There should be option to disable student marking but enable automatic marking only._**
Now you need to check "Allow students to record own attendance" for automatic marking.

**_Will be nice if automarking will check last student activity in the course/site within time frame defined in session settings, not just only log in history._**
For now, if student will log in before session start and keep log in there, auto marking will not mark this student.

gormster:
I'm going to work on the idea of adding as much automatic marking as possible to non-student marking. Here's my roadmap:

- [ ] Get task\auto_mark to mark every student not manually marked by the end of the session with the default value
- [ ] Fix UI so that automatic marking can be enabled and disabled separately to student self-marking
[...]

PR #384, opened 2019-04-24
Fixing Issue #365

jonathanchan2:
Hi Dan, 

I looked into issue #365 and made some changes according to your last suggestion in the issue thread. Please note that I was only able to test my changes on moodle 3.5 since it is the only version of moodle that is currently available to me. 

Hopefully, these changes fully address issue #365.

Best regards,
Jonathan.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/nbproject/private/
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,0 @@
-/nbproject/private/
--- a/add_form.php
+++ b/add_form.php
@@ -189,94 +189,96 @@ class mod_attendance_add_form extends moodleform {
         $mform->setType('previoussessiondate', PARAM_INT);
 
         // Students can mark own attendance.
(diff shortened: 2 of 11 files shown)


jonathanchan2:
I just checked back with this pull request and realised that the travis-ci build failed. The build report says that the "apt-get install failed" and goes onto provide further details: 

"The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install oracle-java8-installer oracle-java8-set-default" failed and exited with 100 during ."

I am faily new to travis-ci and I am not sure why this error would occur. If you have any experience with this error, could you inform me as to how this could be fixed?

danmarsden:
Thanks Jonathan - that's not quite what I was thinking - some of the settings in the attendance settings page like subnets/prevent shared ip etc should be configurable for auto-marking even when student marking is off. I'll see if I can get some time to look at this myself.

Also for future reference - please don't change the version numbers in version.php like that. For more information on how the version numbers work see:
https://docs.moodle.org/dev/version.php

the travis config might need uppdating - I'll take a look at that too. Thanks!

Issue #374, opened 2019-03-15
Problems with position of fixed table row headers for report


Kathrin84:
Hi Dan!

Thank you very much for your new release of this plugin for Moodle 3.6. 

I was testing your plugin and encountered two issues with the reports table.

Currently I'm operating on Moodle 3.6.2+ (Build: 20190222) with mod_attendance 3.6.5
2019022500 and Theme Boost (and Boost Child theme, too).

When editing mode is off, the fixes table row header is placed some pixels too far to the left:
![attendance_report_ediiting_off](https://user-images.githubusercontent.com/4711883/54423251-851f9880-4710-11e9-91c3-b2b147cffa6a.png)

This can be fixed by setting the value left to 20px (https://github.com/danmarsden/moodle-mod_attendance/blob/master/styles.css#L264).

The more severe problem occurs, when editing mode is turned on. The same table looks like this:
![attendance_report_editing_on](https://user-images.githubusercontent.com/4711883/54423543-2dcdf800-4711-11e9-8d9a-f0f85ab23869.png)

The entire fixed table row header is placed right into the content. This is due to this rule, as the editing mode adds .has-region-side-pre:
https://github.com/danmarsden/moodle-mod_attendance/blob/master/styles.css#L279

I've seen that you added these rules to support More and Clean themes: (https://github.com/danmarsden/moodle-mod_attendance/commit/cdf0f331ab6242944511f98b6fadf9a4a25406a1)

Obviously this seems to break displaying in Boost. Even if you titled the commit "improve css for more/clean themes with attendance report." it's not possible to check the selected theme via CSS in Moodle, as far as I know.

It would be great if you could have a look at this!

Thanks in advance and best wishes, Kathrin

danmarsden:
Thanks Kathrin - I thought I'd tested these changes with Boost etc, but obviously not quite enough - there's also a pull request here to deal with multi-line user names that makes me think I might need to redesign the way it floats the user names on the left - unfortunately I've got quite a bit in my backlog at the moment so I'm not sure when I'll get to this but it's definitely high on my list.

thanks!

Syxton:
#388 for your consideration

danmarsden:
thanks to @Syxton for the fix on this one.

PR #388, opened 2019-05-17
Report Boost layout fix

Syxton:


Syxton:
Meant to fix #374

danmarsden:
looks great and works well thanks! - I've just tested with the new classic theme and it looks I might need to do some work there too...

PR #396, opened 2019-06-17
issue#395 - change preventsharedip default value to 1

tuanngocnguyen:


version.php
```diff
@@ -23,7 +23,7 @@
  */
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version  = 2019052801;
+$plugin->version  = 2019061703;
```
danmarsden: this doesn't match the savepoint in the upgrade.php script - looks like you might need to fix that :-)

PR #398, opened 2019-06-20
DBSchema issue

tuanngocnguyen:
Hi Dan,

There are some mismatches between install.xml and upgrade.php

Regards table "attendance", colum "sessiondetailspos"
**Not null in install.xml:**
https://github.com/danmarsden/moodle-mod_attendance/blob/master/db/install.xml#L17
**Null value instead of XMLDB_NOTNULL:**
https://github.com/danmarsden/moodle-mod_attendance/blob/master/db/upgrade.php#L206 

Regards table "attendance_warning", colum "maxwarn":
The default value is missing in install.xm
https://github.com/danmarsden/moodle-mod_attendance/blob/master/db/install.xml#L130
It is '1' in upgrade.php:
https://github.com/danmarsden/moodle-mod_attendance/blob/master/db/upgrade.php#L414


danmarsden:
cool thanks! - merged.

PR #400, opened 2019-06-24
Qr rotation dev

danmarsden:
Creating a pull request for this so we can use in-line comments on the commit


password.php
```diff
@@ -41,6 +41,12 @@
     exit;
 }
 
+if (isset($_GET['returnpasswords']) && $_GET['returnpasswords'] == 1) {
```
danmarsden: raw access to $_GET is not really allowed - can you please replace this with optional_param

lang/en/attendance.php
```diff
@@ -244,6 +244,9 @@
 $string['includeall'] = 'Select all sessions';
 $string['includenottaken'] = 'Include not taken sessions';
 $string['includeqrcode'] = 'Include QR code';
+$string['rotateqrcode'] = 'Rotate QR code';
+$string['rotateqrcodeinterval'] = 'Rotate QR code interval';
+$string['rotateqrcodeinterval_desc'] = 'Time interval to rotate QR codes by.';
```
danmarsden: It is not clear what an "interval" is - it might be better to replace this with something like "Rotate QR code interval (seconds)" or "The time in seconds to rotate the QR code."

attendance.php
```diff
@@ -37,6 +37,33 @@
...
+    // Check if cookie is set and verify
+    if (isset($_COOKIE[$cookiename])) {
+        // Check the token
+        if ($secrethash !== $_COOKIE[$cookiename]) {
+            // Flag error
+            print_error('qr_cookie_error', 'mod_attendance', $url);
+        }
+    } else {
+        // Check password
+        $qrpassdatabase = $DB->get_record_sql('SELECT * FROM {attendance_rotate_passwords} WHERE attendanceid = ? AND expirytime > ? ORDER BY expirytime ASC LIMIT 1', ['attendanceid'=>$id, time()], $strictness=IGNORE_MISSING);
```
danmarsden: I'd like this to also accept the "previous" password if it's expiry was less than 2 seconds ago - hard code it at 2 seconds for now, but we can look at adding a site level setting to control this - it means that if the server is running slow and takes a bit long to receive the request, the code is still accepted - does that make sense?

settings.php
```diff
@@ -46,6 +46,10 @@
          1000 => 1000,
     );
 
+    $settings->add(new admin_setting_configtext('attendance/rotateqrcodeinterval',
```
danmarsden: can you please drop this under the studentscanmark setting as it relates to that functionality - thanks!

danmarsden:
also looks like a bunch of travis failures that will need to be fixed too.

maksudr:
@danmarsden I have actioned the requested changes.

classes/task/clear_temporary_passwords.php
```diff
@@ -0,0 +1,53 @@
...
+        return get_string('rotateqrcode_cleartemppass_task', 'mod_attendance');
+    }
+
+    /**
+     * Execute the task.
+     */
+    public function execute() {
+        global $DB;
+
+        $DB->delete_records('attendance_rotate_passwords');
```
danmarsden: shouldn't this only be deleting passwords with an old expirytime ? - some weird behaviour could occur if it deletes passwords while an attendance is being used! :-)

PR #406, opened 2019-07-10
Fix #405

### LINKED ISSUE #405: Too easy to accidentally override existing marks with 'set status for all users' radiobutton.
We've found the "set status for all users" radiobuttons to be both essential and infuriating. It is normal for some students' statuses to be set to a state that indicates a "excused absence" in advance of a session, which means the "set status for all users" cannot be used to mark all those who are expected to be present as present at the time of the session.

We've implemented a fix so that the "set status for all users" will only set status for users whose status is not already set, unless shift-clicked, in which case it will set for all users and override existing statuses.

nwp90:
Click to set all only sets rows that are currently unset; shift-click to actually set all.

### DIFF
--- a/renderer.php
+++ b/renderer.php
@@ -663,7 +663,12 @@ class mod_attendance_renderer extends plugin_renderer_base {
             $PAGE->requires->js_amd_inline("
                 require(['jquery'], function($) {
                     $('#radiocheckstatus".$st->id."').click(function(e) {
-                        $('#attendancetakeform').find('.st".$st->id."').prop('checked', true);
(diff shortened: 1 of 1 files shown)


nwp90:
Haven't fixed tests yet; happy to do so if you agree with basic aim.

danmarsden:
I'm closing this as a pull request as I don't think the approach is quite right for UX, but I'm hoping maksudr might take your code and make it more user friendly - hopefully we'll see a new Pull request from him soon! :-)

PR #408, opened 2019-07-15
Show attendance events on Timeline Block

noisyjerm:
This will allow attendance sessions with calendar events to be displayed on the Moodle 3.7 timeline block

Screen Shot 2019-07-15 at 2 16 09 PM


danmarsden:
thanks for the pull request - sounds like a good idea - the travis checks here are reporting some coding violations and some issues with sql statements - looks like some invalid double quotes used in the sql? - any chance you could take a closer look at that to see if you can fix that up?

noisyjerm:
Hi Dan
I've addressed the coding violations, just need to merge in the last 3 commits from master next.

lib.php
```diff
@@ -529,4 +529,62 @@ function (stdClass $warning) use ($userid) : bool {
...
+    $sess = $DB->get_record_sql($sql, [$event->id]);
+
+    $context = context_module::instance($sess->cmid);
+
+    if (has_capability('mod/attendance:takeattendances', $context) && $sess->sessdate < time()) {
+        $actiontext = get_string('takeattendance', 'mod_attendance');
+        $url = '/mod/attendance/take.php';
+        $param = ['id' => $sess->cmid, 'sessionid' => $sess->id, 'grouptype' => 0];
+    } else if (has_capability('mod/attendance:canbelisted', $context)) {
+        $sql = 'SELECT s.description FROM {attendance_log} l '
```
danmarsden: this looks very wrong - it should be using the function attendance_can_student_mark()

classes/calendar_helpers.php
```diff
@@ -50,13 +50,14 @@ function attendance_create_calendar_event(&$session) {
     $caleventdata->courseid       = $attendance->course;
     $caleventdata->groupid        = $session->groupid;
     $caleventdata->instance       = $session->attendanceid;
-    $caleventdata->timestart      = $session->sessdate;
+    $caleventdata->timestart      = $caleventdata->timesort       = $session->sessdate;
```
danmarsden: I'd prefer you match the surrounding code here - one line per variable - can you please drop this onto another line?

classes/calendar_helpers.php
```diff
@@ -135,9 +136,10 @@ function attendance_update_calendar_event($session) {
     // Boring update.
     $caleventdata = new stdClass();
     $caleventdata->timeduration   = $timeduration;
-    $caleventdata->timestart      = $timestart;
+    $caleventdata->timestart      = $caleventdata->timesort       = $timestart;
```
danmarsden: please match the surrounding code (one line per variable) - thanks!

.travis.yml
```diff
@@ -47,4 +47,4 @@ script:
   - moodle-plugin-ci grunt
   - moodle-plugin-ci phpdoc
   - moodle-plugin-ci phpunit
-  - moodle-plugin-ci behat
\ No newline at end of file
+  - moodle-plugin-ci behat
```
danmarsden: can you please remove this carriage return change? - Ideally a patch should only touch files that are associated with the change. thanks!

db/upgrade.php
```diff
@@ -628,5 +628,17 @@ function xmldb_attendance_upgrade($oldversion=0) {
 
     }
 
+    if ($oldversion < 2019062300) {
+        // make entries in (calendar) event table compatible with timeline block
+        $lookbackdays = time() - 30 * (60 * 60 * 24); // no point changing all historical events on large sites
```
danmarsden: This feels a little arbitrary to me.. although do we even need to deal with historical events? - the timeline block only shows "future" events right? - maybe we don't need to convert events older than time() at all?

lib.php
```diff
@@ -529,4 +529,62 @@ function (stdClass $warning) use ($userid) : bool {
...
+ *
+ * This is used by block_timeline in order to display the event appropriately. If null is returned then the event
+ * is not displayed on the block.
+ *
+ * @param calendar_event $event
+ * @param \core_calendar\action_factory $factory
+ * @param int $userid
+ * @return \core_calendar\local\event\entities\action_interface|\core_calendar\local\event\value_objects\action
+ */
+function mod_attendance_core_calendar_provide_event_action(calendar_event $event,
```
danmarsden: The implementation of this function appears to be a bit different from core which uses modinfo and checks uservisible and makes capability checks before doing any sql queries - I also see some use in core functions of global $USER when userid = 0

ideally we should be writing unit tests for this change too - it appears that this stuff is quite easy to test and there are a number of examples of unit tests for existing plugins that could be used

danmarsden:
Thansk @noisyjerm - I've made a few comments inline above - hopefully most should be easy to fix and you can include some unit tests for this as well!

db/upgrade.php
noisyjerm: Agree the 30 days is arbitrary, perhaps a setting would be more appropriate? The timeline block can show "Overdue" events. I see a use case for teachers who have not marked the attendance for previous sessions.

lib.php
noisyjerm: what i am trying to do here (line 567) is say "are you are a student". The timeline event / session should always show to the student but the action link will change depending.

If attendance has been taken (line 571) the status is shown. If the attendance has not been taken and the session is in the past then perhaps i should be using attendance_can_student_mark().

If they can mark their own attendance, the link to take.php should be shown (line 578).

Future sessions do not show the action link. 

Comments included below
`if (has_capability('mod/attendance:takeattendances', $context) && $sess->sessdate < time()) {
        // The user can take attendance and the session is in the past. Show link to take (take.php)     
        $actiontext = get_string('takeattendance', 'mod_attendance');
        $url = '/mod/attendance/take.php';
        $param = ['id' => $sess->cmid, 'sessionid' => $sess->id, 'grouptype' => 0];
    } else if (has_capability('mod/attendance:canbelisted', $context)) {
        // The user can have their attendance taken
        $sql = 'SELECT s.description FROM {attendance_log} l '
             . 'JOIN {attendance_statuses} s ON s.id = l.statusid '
             . 'WHERE sessionid = ? AND studentid = ?';
        $recorded = $DB->get_record_sql($sql, [$sess->id, $userid]);

        if ($recorded !== false) {
            //  Attendance has been taken. Show the status and link to the attendance report (view.php)
            $actiontext = $recorded->description;
            $url = '/mod/attendance/view.php';
            $param = ['id' => $sess->cmid];
        } else {
            // if student can mark own attendance and the session is in the past show link to take (attendance.php)
            $actiontext = $sess->studentscanmark && $sess->sessdate < time()
                        ? get_string('submitattendance', 'mod_attendance') : "";
            $param = ['sessid' => $sess->id];
        }
    }`

danmarsden: yeah - the action link to take attendance for a student should only show when attendance_can_student_mark returns that it can be shown - take a look at the other examples where that function is used. You can still show the content of the event, but you should not be displaying a link to take.php unless attendance_can_student_mark says that the user can actually take attendance for that session.

db/upgrade.php
danmarsden: right - forgot about overdue events - using a setting to control upgrade scripts doesn't usually work too well, I'm ok with the arbitrary decision to only care about events older than 30 days for now - although it would be nice for you to also make sure the coding guideline warnings that will be showing up for those comments (and any other new changes) are also fixed and not just the errors... (check the travis log for all the new warnings that your code is creating) - this is particularly useful because the moodle.org plugins db reports on them too and I'm down to 6 warnings in the last version I pushed through.

one of the items that it will be picking up is that your comments should be starting with a capital letter and ending in a full stop... (annoying, but it does make the code clean when it follows guidelines!) :-)

danmarsden:
I'm closing this pull request for now as it still needs some more work before it can be merged - but I've added #429 in the issues tracker to remind about it.

Feel free to resubmit another pull request after you have fixed the items requested - thanks!

Issue #412, opened 2019-07-23
log entry for report viewed generates incorrect URL


nwp90:
Log entries for "report viewed" events generate an incorrect URL, as the attendance instance id is passed to \mod_attendance\event\report_viewed::create as objectid, but report.php expects id to be a cmid.

I think report_viewed.php's get_url just needs to use this->contextinstanceid instead of this->id to generate the url. But *all existing* report_viewed log entries should probably be fixed too.

danmarsden:
just fix it without changing the historical logs... making update calls on the logs tables makes me very nervous with the number of possible records that could be affected, and the fact that no-one has noticed this until now...

nwp90:
OK, happy to do that. Wasn't so keen if all the updating was going to happen ;)

danmarsden:
fixed in #413 - thanks!

PR #413, opened 2019-07-24
Fix logs generated for report_viewed event to include correct id param.

nwp90:
Fix for #412. Fixes new and legacy log entry generation. Not updating old log entries.
I am not currently able to test with legacy logs.

### DIFF
--- a/classes/event/report_viewed.php
+++ b/classes/event/report_viewed.php
@@ -74,7 +74,7 @@ class report_viewed extends \core\event\base {
      * @return \moodle_url
      */
     public function get_url() {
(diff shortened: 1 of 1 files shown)


nwp90:
Ah bother. You're too quick. The legacy fix is wrong; I misread what it was doing. Looks like it needs a cmid.

nwp90:
Just wrote and then deleted a comment here - I think the legacy fix is wrong; according to add_to_log in lib/deprecatedlib.php, that param should be a cmid. However, looking through mod/*/classes/event/*, I only ever see contextinstanceid being used as I have here.

Let's all be wrong together.

danmarsden:
haha - cool :-)

PR #414, opened 2019-07-25
QR rotation dev - timer, rotate text password, expiry time margin setting

maksudr:
- Adding expiry time margin to setting and updating settings text.
- Add rotateqrcode to class structure.
- Add submit password text to lang.
- Form to allow users to enter rotating passwords.
- Display text password on password.php
- Add timer for rotate password/qrcode on password.php


danmarsden:
Thanks @maksudr - I like the password entry in the view.php instead of having to click on "submit attendance" - although I do think we need to tidy up the 2nd page - we shouldn't present the internal text password at all, and if "Automatically select highest status available" is selected we should just save the attendance rather than presenting them with a form to submit. I allso think we'll need to do some css work on the password page to make it look a little tidier but I've merged this into the master branch. (I've just split off new stable branches for Moodle 3.6/3.7) to make "master" more of an in-development branch.

PR #409, opened 2019-07-17
More webservice functions.

kabalin:
This PR contains new webservice functions, namely:

* `mod_wsattendance_add_attendance`
* `mod_wsattendance_remove_attendance`
* `mod_wsattendance_add_session`
* `mod_wsattendance_remove_session`

They are designed to facilitate remote session creation in courses. All functions contain permission validation. Also basic permission check has been added to all existing webservice functions. Unit test coverage for all new functions has been added as well.

As side note, I think it might be good idea to rename functions (no idea about impact) to match common naming convention (e.g. `mod_attendance_*` instead of `mod_wsattendance*`). I can add commit if needed.

The work on those changes has been funded by WKS KV Bildung (https://www.wksbern.ch).

PR #416, opened 2019-07-29
Rename webservices to match Moodle naming convention.

kabalin:
Hi @danmarsden 😃 

WS functions renaming as discussed in #409


danmarsden:
great thanks!!

PR #417, opened 2019-07-30
Carefulclickall - Drop Down - Fixes #405

### LINKED ISSUE #405: Too easy to accidentally override existing marks with 'set status for all users' radiobutton.

maksudr:
Building on the idea in #405. Converting shift-click functionality into a dropdown.

@danmarsden I have made the UX changes discussed. Let me know what you think!


maksudr:
dropdown1
dropdown2

danmarsden:
Sorry - I thought I'd responded to this last week! yeah - that's more like it I think, although the selected value should be stored as a user preference so that when viewing the page, the users preferred option is always selected.

 @nwp90 what do you think?

(obviously we'd need to adjust some css here, but I think it's in the right location and it feels more logical.)

nwp90:
Doesn't feel 100%, but it'll do the job.

One thing that bugs me slightly is that it seems to me that the more destructive option should always be the one requiring extra effort to achieve.
 
I've not done much mobile stuff so I don't know whether there's a "standard" way of doing this kind of thing (yet?)...

renderer.php
```diff
@@ -649,7 +652,21 @@ protected function render_attendance_take_list(attendance_take_data $takedata) {
         foreach ($extrasearchfields as $field) {
             $row->cells[] = '';
         }
-        $row->cells[] = html_writer::div(get_string('setallstatuses', 'attendance'), 'setallstatuses');
+
+        if (get_user_preferences('mod_attendance_statusdropdown') == 'unselected') {
```
danmarsden: might be good to drop this into a helper function as it's used twice in the code - we should also be using lang strings for the "all" and "unselected" items shouldn't we?

maksudr:
@danmarsden I'verefactored the code with a helper function. Should be good to go :+1:

danmarsden:
merged this in manually! - thanks guys!! :-)

Issue #432, opened 2019-09-21
RTL user interface - The locations of the hour and minutes dropdowns are incorrectly reversed


acquaalta:
Hi,
Thank you for this wonderful plugin.

It seems that in an RTL (Right-to-left) environment, the timing UI is the exact opposite of the the one in the LTR environment. Please see attached a relevant screenshot.
![Attendence - Hours and Minutes - RTL vs LTR](https://user-images.githubusercontent.com/51836675/65372845-ee630400-dc7d-11e9-8406-ef5b466833e5.png)

Even in RTL environments, we still use the time format in the same way as in LTR environments - we read the time from left to right. Is it possible to swap the locations of the hour and minutes dropdowns in each relevant segment of the UI?

Thanks in advance :-)

danmarsden:
thanks for the report, feel free to help us work on this issue.

danmarsden:
thanks to @nadavkav a fix for this is now in the 3.5 and higher branches here on github.

PR #434, opened 2019-10-02
Swap hour:min input fields in a new session form, when in RTL

nadavkav:
Should fix issue #432

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -177,12 +177,21 @@ function attendance_form_sessiondate_selector (MoodleQuickForm $mform) {
     }
 
     $sesendtime = array();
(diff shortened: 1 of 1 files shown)


danmarsden:
thanks @nadavkav - it does look a little weird like that but I couldn't see a cleaner way to write it without putting a lot more effort in to rewrite the way the element works - have merged it into 3.5 and higher branches!

nadavkav:
@danmarsden thank you for reviewing it and merging it.
Indeed, I was also looking for a cleaner way to fix it, but did not want to introduce too much code changes.

danmarsden:
yeah - I almost rewrote it so that it didn't duplicate the static vars, but that just looked even worse! :-)

PR #435, opened 2019-10-04
Upgrade.php fix to sync with install.xml

Peterburnett:
FIxes fields where Upgrade.php sets values differently to install.xml, resolves problems from existing installs when running admin/cli/check_database_schema

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -628,5 +628,32 @@ function xmldb_attendance_upgrade($oldversion=0) {
 
     }
 
+    if ($oldversion < 2019080802) {
+
+        // Changing nullability of field sessiondetailspos on table attendance to not null.
+        $table = new xmldb_table('attendance');
+        $field = new xmldb_field('sessiondetailspos', XMLDB_TYPE_CHAR, '5', null, XMLDB_NOTNULL, null, 'left', 'subnet');
+
+        // Launch change of nullability for field sessiondetailspos.
+        $dbman->change_field_notnull($table, $field);
+
+        // Changing the default of field preventsharedip on table attendance_sessions to 1.
+        $table = new xmldb_table('attendance_sessions');
+        $field = new xmldb_field('preventsharedip', XMLDB_TYPE_INTEGER, '1', null, XMLDB_NOTNULL, null, '1', 'absenteereport');
+
+        // Launch change of default for field preventsharedip.
+        $dbman->change_field_default($table, $field);
+
+        // Changing the default of field maxwarn on table attendance_warning to drop it.
+        $table = new xmldb_table('attendance_warning');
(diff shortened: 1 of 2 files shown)


db/upgrade.php
```diff
@@ -628,5 +628,32 @@ function xmldb_attendance_upgrade($oldversion=0) {
...
+        // Changing the default of field preventsharedip on table attendance_sessions to 1.
+        $table = new xmldb_table('attendance_sessions');
+        $field = new xmldb_field('preventsharedip', XMLDB_TYPE_INTEGER, '1', null, XMLDB_NOTNULL, null, '1', 'absenteereport');
+
+        // Launch change of default for field preventsharedip.
+        $dbman->change_field_default($table, $field);
+
+        // Changing the default of field maxwarn on table attendance_warning to drop it.
+        $table = new xmldb_table('attendance_warning');
+        $field = new xmldb_field('maxwarn', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, null, null, 'warnafter');
```
danmarsden: This looks weird? - the default was incorrectly set at some point, but it was fixed in an upgrade step and the default is different here from what is defined in install.xml.

I wonder if the site you're working on has been upgraded incorrectly at some point.

db/upgrade.php
```diff
@@ -628,5 +628,32 @@ function xmldb_attendance_upgrade($oldversion=0) {
...
+        // Changing nullability of field sessiondetailspos on table attendance to not null.
+        $table = new xmldb_table('attendance');
+        $field = new xmldb_field('sessiondetailspos', XMLDB_TYPE_CHAR, '5', null, XMLDB_NOTNULL, null, 'left', 'subnet');
+
+        // Launch change of nullability for field sessiondetailspos.
+        $dbman->change_field_notnull($table, $field);
+
+        // Changing the default of field preventsharedip on table attendance_sessions to 1.
+        $table = new xmldb_table('attendance_sessions');
+        $field = new xmldb_field('preventsharedip', XMLDB_TYPE_INTEGER, '1', null, XMLDB_NOTNULL, null, '1', 'absenteereport');
```
danmarsden: the install.xml file states a default of 0 for preventsharedip - sounds like you have a weird install.xml file?

danmarsden:
I'm closing this as it looks invalid and doesn't seem to be consistent with the latest code. I suspect it's due to some old code used on your site. Once you have checked that you are using the latest code, and have verified there is still an issue feel free to update your pull reuqest with the correct changes. Thanks!

Peterburnett:
Apologies for Friday brain, this patch was submitted against the wrong branch. I have created a new PR for the correct branch here.
https://github.com/danmarsden/moodle-mod_attendance/pull/436

PR #436, opened 2019-10-07
Fixed DB schema issues where upgrade.php doesnt align with install.xml

Peterburnett:
The site is currently using this branch, on a Moodle 3.5 installation. I have verified that the issues are present in this branch of the repository. e.g [L17](https://github.com/danmarsden/moodle-mod_attendance/blob/MOODLE_35_STABLE/db/install.xml#L17) of install.xml conflicts with [L206](https://github.com/danmarsden/moodle-mod_attendance/blob/MOODLE_35_STABLE/db/upgrade.php#L206) of upgrade.php

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -550,5 +550,32 @@ function xmldb_attendance_upgrade($oldversion=0) {
         upgrade_mod_savepoint(true, 2018051408, 'attendance');
     }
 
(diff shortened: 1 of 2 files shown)


danmarsden:
This still doesn't look right. It would cause inconsistencies when someone upgrades to a newer release. You would be better to backport the fix already in the newer branches

Issue #352, opened 2018-11-29
Provide link to all course attendance report on users profile page.


danmarsden:
To view the "All courses" attendance report you must be inside an existing attendance, click on the user and then hit the "all courses" tab. It might be useful to format this report in a way that allows us to provide a link directly on a users profile page under "reports" rather than having to enter an existing attendance first.

also mentioned here:
https://moodle.org/mod/forum/discuss.php?d=379137#p1528615

PR #458, opened 2020-03-24
Provide link to all courses attendance report on users profile page

ShivCharanSharma:
This pull request is for #352 .It  provide a link of all courses attendance directly on a users profile page under "reports".


danmarsden:
thanks for the pull request! - unfortunately this isn't quite the full picture - we need to modify the "allcourses" report to allow just a courseid to be passed, and there needs to be some better capability checking to make sure the user can access the report before showing the link.

Your code also fails a bunch of our automated tests.

Feel free to improve your pull request, otherwise hopefully one day someone else might give this a go!

Issue #443, opened 2019-11-19
Report page doesn't work properly with Clean and Boost themes


Urpokarhu1:
Hello,

Report page doesn't work properly with Clean and Boost themes. 

With Clean there is a weird white-row between user name and email address.
With Boost usernames overlap email address.

![boost](https://user-images.githubusercontent.com/53904029/69125515-a3fbd900-0aae-11ea-9470-4f73d5535dcf.PNG)

![clean](https://user-images.githubusercontent.com/53904029/69125529-aa8a5080-0aae-11ea-9f06-c934deb2f9a3.png)

danmarsden:
can you please clarify which version of Moodle you are using and if you are using the latest version of the code from github, or the much older release from the moodle.org plugins database?

Urpokarhu1:
Hello,

Moodle 3.5 and version from moodle plugins page:

https://moodle.org/plugins/report/index.php?report=user_sites&pluginid=772

danmarsden:
The lastest code is always here in the github repo.

I'm pretty sure this is a duplicate of #388 but it doesn't look like that was backported to the 3.5 branch - feel free to help by submitting a pull request or testing the changes in #388 on your site.

NeillM:
We are seeing this in Moodle 3.8

It looks like the rules for making it work in the old clean theme might be the cause, i.e.

#page-mod-attendance-report.has-region-side-pre:not(.empty-region-side-pre) .attendancereporttable .headcol

I could see this rule is still in the master branch.

danmarsden:
Thanks Neil, feel free to help to fix by submitting a pull request :-)

NeillM:
Pull request with a suggested fix sent.

danmarsden:
this one should be resolved now - but please report any further issues as a new bug including a screenshot - thanks!

PR #469, opened 2020-05-11
Fix styling of floating username column in reports

NeillM:
This will fix the display issue in #443 under Classic and Boost

Not 100% certain if it will work perfectly with old bootstrapbase themes such as Clean though (since I do not have one with Moodle 3.8) Do they even still need supporting?

Making the main div relative means that the header column is absolute to that (and not the whole of the page) so blocks no longer affect where it is positioned,

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -251,6 +251,10 @@
     }
 }
 
+#page-mod-attendance-report div[role=main] {
+    position: relative;
+}
+
 #page-mod-attendance-report .attendancereporttable {
     overflow-x: scroll;
     overflow-y: visible;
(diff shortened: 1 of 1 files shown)


danmarsden:
fyi - this got closed "automatically" as I renamed the "master" branch to "main" - I'll see if there's a way I can reopen this soon.

danmarsden:
cherry-picked this one manually into the new main branch - thanks Neil!

NeillM:
I did wonder why it was closed with no explanation, that makes sense though!

danmarsden:
I didn't even notice at the time!!! Went looking for something and couldnt figure out why they had disappeared!!

PR #404, opened 2019-07-09
allsessionsreport

nwp90:
Partial fix for #373, but actually motivated by our own internal needs.

- The intent is to add a report on a single student across the entire Moodle instance.
- I've tried to keep changes to existing code minimal, but I think in a few places a little tweaking was necessary.
- I've kept what I've added (painfully) straightforward for now; once/assuming it gets merged, IIRC a fair bit of refactoring would make sense.
- In particular, even though the desired output is different, the structure of the query in get_user_sessions_log_full (added here), as opposed to the two queries in (the existing) get_user_sessions_log_extended should be significantly faster and I think more correct (the existing function does not get all the sessions it implies it should).
- The new report is fairly wide - it may be that things will need tweaking to work well with small screens or certain themes.

PR #475, opened 2020-06-17
Nwp90 allsessionsreport

danmarsden:
creating a new pull request against the correct branch so I don't forget about this one.


lang/en/attendance.php
```diff
@@ -431,7 +434,12 @@
 $string['sessiongenerated'] = 'One session was successfully generated';
 $string['sessionunknowngroup'] = 'A session specifies unknown group(s): {$a}';
 $string['sessions'] = 'Sessions';
+$string['sessionsallcourses'] = 'All courses';
+$string['sessionsbyactivity'] = 'Attendance Instance';
```
danmarsden, line 438: Camel case - should be lower case i

lang/en/attendance.php
```diff
@@ -546,6 +556,7 @@
 $string['warningpercent_help'] = 'A warning will be triggered when the overall percentage falls below this number.';
 $string['warningthreshold'] = 'Warning threshold';
 $string['week'] = 'week(s)';
+$string['weekcommencing'] = 'Week Commencing';
```
danmarsden, line 559: Camel case.

danmarsden:
just adding a note for me to come back to...

need to review/test security access of new report prior to merge.

danmarsden:
<b>Notice</b>:  Undefined index: sesscourses in <b>/home/dan/www/moodle/mod/attendance/classes/event/session_report_viewed.php</b> on line <b>85</b><br />

danmarsden:
Notice: Undefined offset: 1 in /home/dan/www/moodle/mod/attendance/renderer.php on line 1975

danmarsden:
Hey Nick - finally got a chance to take a look at this - will be great to see this land!! - took a bit to find the "turn editing on" function was needed to be able to edit the attendance values in the all sessions report, and there are some css issues with that - not sure if it relates to the other upstream changes or if there's some other css missing? - I'll add a screenshot from the boost theme that shows you what I mean.

Is this something you have time to update? - no worries if not, I'll try to come back to it myself.

danmarsden:
editingon

danmarsden:
pinging @nwp90 just in case the above comments didn't ping you! :-)

danmarsden:
we should also find a way to drop the "turn editing on" button into the all sessions report page so that it can be turned on/off when viewing the all sessions report.

nwp90:
I’d love to have a chat about your thoughts on this in general – happy to do the work to get it into what you see as a “fit state” once I know that you’re at least keen on what I’m trying to achieve.

Cheers,


Nick

danmarsden:
Would be really happy to see this land - had another email from someone this past week that wanted this as well!  the other thing that would be good to modify is the reflectionclass - lets just make attendance_take_data a public method?

danmarsden:
I'd like to look at rewriting a bunch of those reporting pages when Moodle 4.0 ends up with the new reporting API's but that's a long way away... :-)

nwp90:
Yes, will likely be working on this again RSN. Comments indeed either didn't ping or I missed it. Just came here as I've had the screenshotted problem reported in our internal tracker too :)

nwp90:
Re-familiarising myself now...

nwp90:
Line 1975 in renderer.php doesn't seem to have any way of generating an undefined offset in what I'm looking at - what is the line that's generating that?

nwp90:
Just copying some of the comments from the previous PR (#404) in here:

-     The intent is to add a report on a single student across the entire Moodle instance.
-     I've tried to keep changes to existing code minimal, but I think in a few places a little tweaking was necessary.
-     I've kept what I've added (painfully) straightforward for now; once/assuming it gets merged, IIRC a fair bit of refactoring would make sense.
-     In particular, even though the desired output is different, the structure of the query in get_user_sessions_log_full (added here), as opposed to the two queries in (the existing) get_user_sessions_log_extended should be significantly faster and I think more correct (the existing function does not get all the sessions it implies it should).
-     The new report is fairly wide - it may be that things will need tweaking to work well with small screens or certain themes.

nwp90:
On attendance_take_data -- will have to allow constructor not to do all the work when it's not needed - maybe if the structure is not passed in then don't set anything up?

nwp90:
Pushed the trivial fixes.

nwp90:
Another thing that will need looking at is where travis is complaining about use of error_log - is that the right permission to be checking (I wasn't clear what the intent of takeattendances vs manageattendances and changeattendances was, IIRC), and how to notify user of failed permission check.

danmarsden:
Haven't looked at the code properly - but surely a require_capability can be used in there - eg it should stop processing if someone is doing something they don't have permissions for...

nwp90:
Yep, OK, but I wasn't clear what the intent of the various capabilities is... which one(s) should it be checking?

danmarsden:
looks like I forgot to respond there - the "takeattendnace" one you're checking there is correct - but it should be checking "changeattendance" around here:
https://github.com/danmarsden/moodle-mod_attendance/pull/475/files#diff-2bfd735f11fe3f1baf0ae268d3d17dfdR629

basically - some users are allowed to set the intitial attendance, but then aren't trusted enough to "update" an existing attendance record.

But - some of this really needs to be using generic/shared functions - eg - you've copied the "take_sessions_from_form_data" partly from structure.php take_from_form_data which I've recently restructured into 2 parts (split some of it into a new "save_log" which is also used by csv session marking - I suspect we might be able to improve the core take_from_form_data() to work in your form as well as the other forms.

Let me know if you want to give a go at modifying the core take_from_form_data to also accept the data passed from your new form (with extra checks on permissions etc.) or if you want to "hand-over" and get me to tidy up the rest! :-)

nwp90:
Agree re integration - I was trying to write it with minimal modifications to existing code in case we ended up having to maintain it as a patch. If you're happy to tidy it up, go nuts. Feel free to ping me if there's anything that doesn't make sense, or if you need a hand.

danmarsden:
finally got a chance to merge a modifed version of this into the main branch.

I've created two follow up issues #505 and #506 to deal with the duplicated code/nasty hack

also dropped in a github actions file at the same time - stole the config from kabalin's repo :-)

nwp90:
Round of applause... 🥇  ;)

PR #485, opened 2020-07-29
Add caleventid index to attendance_sessions table

kristian-94:
@danmarsden  Do you want to give this a review? I've just fixed for the 37 branch which is the one affecting me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ services:
   - mysql
 
 addons:
-  firefox: "47.0.1"
-  postgresql: "9.4"
+  postgresql: "9.5"
   apt:
     packages:
       - openjdk-8-jre-headless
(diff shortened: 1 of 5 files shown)


danmarsden:
Thanks @kristian-94 - we'll need to drop this into the main branch first and then backport to the others as I can't see it being added anywhere in the upgrade scripts. I'll get to this at some point unless you beat me to it with a pull request! :-)

kristian-94:
@danmarsden  We're getting travis failures on moodle 37 but at least now with these changes they're the same failures as are already there

danmarsden:
I've merged it into the main branch now, but ... 
https://github.com/danmarsden/moodle-mod_attendance/pull/485/files#diff-5bbd3ccd25214f9956eea7e9f714bc08L26

stable version should only bump by 1 - we only use the current date in the main branch as per rules in:
https://docs.moodle.org/dev/version.php

unfortunately that will also mean anywhere you've deployed that patch on a 3.7 branch will need a hack to downgrade the attendance version. (hopefully you haven't dropped it into many sites yet.)

kristian-94:
@danmarsden  That's fine, I've only done testing on my local machine anyway so far. I've updated the version to be 2019072403 now

danmarsden:
Great! - thanks for that, I've merged it in. accidentally did squash/merge (default github) - might have been better to merge instead of squash though.

PR #476, opened 2020-06-17
Upload Attendance Feature

danmarsden:
Creating a new pull request so we don't forget about this one.

See previous pull requests #393 and #383 for the previous history.

PR #486, opened 2020-07-30
Feature upload

danmarsden:
rebase of pull #476


danmarsden:
taking an internal note to remind me...
take_from_form_data changes need to be tidied up - $data shouldn't be from raw form and also mod_attendance_importer class.

danmarsden:
@jonathanchan2 - I'm finally coming back to this again and have started tidying the patch up a bit.

I don't like the 2 separate columns "scandate" and "scantime" and would prefer a single column that contains the date/time. Does the scanner you're using separate these out completely? - or does the scantime include a full datestamp?

Any chance you could fire me through a raw csv file with a handful of rows that shows the export you get that comes directly from the scanner you guys are using?

thanks,

danmarsden:
I've tided this up and merged in manually. thanks!

PR #491, opened 2020-08-21
Fix #490 add admin setting for admin to control the fields for "Ident…

### LINKED ISSUE #490: add admin setting for admin to control the fields for "Identify student by"
Add an admin setting where an admin can control (with a multiselect box) which user profile fields will be added to the export table. This admin setting will offer the possibility to check / uncheck a) the internal Moodle user id b) all core profile fields c) all custom profile fields.
• The export feature will respect these settings. The teacher will just be able to export the fields which are enabled by the admin. Additionally, the teacher can disable some fields to reduce the table (just like he is able currently).
• The only field which will be exported as a mandatory field is the real name of the user.
• The export feature will respect the visibility of the fields based on the corresponding access control mechanisms. If a teacher is not allowed to see the value of a particular core profile field or custom profile field, then the export feature has to respect that.

yao9394:
…ify student by"

### DIFF
--- a/classes/structure.php
+++ b/classes/structure.php
@@ -801,7 +801,7 @@ class mod_attendance_structure {
     public function get_users($groupid = 0, $page = 1) : array {
         global $DB;
 
-        $fields = array('username' , 'idnumber' , 'institution' , 'department');
+        $fields = array('username' , 'idnumber' , 'institution' , 'department', 'city', 'country');
         // Get user identity fields if required - doesn't return original $fields array.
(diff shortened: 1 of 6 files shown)


danmarsden:
@yao9394 - looks like the behat tests for the export with extra user fields is failing - also some coding guideline failures with the new code - can you please take a look at that?

the other failure looks like a new requirement for postgres 9.6 in 3.10 - I'll fix that in the main config now

PR #499, opened 2020-10-25
Issue496 - mod_attendance_generator should return stdClass - fix unit tests in 3.9

kristian-94:
I had a bunch of failing unit tests in 3.9 because of this. This is what I did to fix it.


alexmorrisnz:
@danmarsden we have a few sites failing unit tests because of this. I've put this patch on a demo site and it fixed the tests.

danmarsden:
Thanks for the PR Kristian and to Alex for the review - merged!

PR #526, opened 2021-05-03
Feature/ccle 9110 attendance module render issue

deanjones-git:
We fixed the issue with long student names overflowing as below:
image

### DIFF
--- a/classes/structure.php
+++ b/classes/structure.php
@@ -653,7 +653,11 @@ class mod_attendance_structure {
         $record->studentid = $USER->id;
         $record->statusid = $mformdata->status;
         $record->statusset = $statuses;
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks @deanjones-git - this looks very close!

unfortunately the addition of "background: white;" in the headcol breaks the alternating row colour backgrounds, so we'd need to find another way to fix that.

You have also got another unrelated commit in this PR and I can't merge this with the "UCLA" style comments so if you could remove them as well it would be great!

thanks!

danmarsden:
Although - I just realised that we already break the alternating background colours in the existing version, so I might just grab your changes and merge it manually.

danmarsden:
merged this one manually - thanks for the code!

PR #535, opened 2021-06-25
Adapt mobile app code to Ionic 5

dpalou:
Hi Dan,

here's the PR to adapt the mobile app code to Ionic 5. I did some tests, but it would be great if you could test it too, just in case.


danmarsden:
Looks great thanks @dpalou !!

Issue #543, opened 2021-08-02
QR Code self marking - error messages


kimberley-johnston:
Hi,

We've been using the attendance mod (thank you!) and have recently been trialling using QR codes to have our students self mark their group sessions. We've run into a couple of teething issues and one of them seems to be if a student attempts to scan into a group session when they're not a member of the group. Rather than getting a failure message they get the success message and no attendance is recorded. 

Is it possible to catch that particular error and show a different message for the students? 

We're currently using Moodle 3.10.5+ with Attendance 3.9.3

danmarsden:
thanks for the report - that does sound like a bug! - I'm quite busy at the moment and don't have a lot of volunteer time to spread across all the things at the moment - but if you have internal development capabilities feel free to submit a pull request, otherwise I'll try to get to this at some point!

thanks,

kimberley-johnston:
Cheers Dan - I'll see what I can do re: internal development.

nwp90:
Working on it...

PR #544, opened 2021-08-03
Only allow successful QR checkin if logged-in user was supposed to be in session

nwp90:
If a student wasn't supposed to be in a session, they've probably scanned
the wrong QR code and it's not helpful to give them a "success" message.


nwp90:
Fixes #543

danmarsden:
finally got round to fixing this - now in the latest version, used a slightly different method than your patch in the end - thanks!

PR #558, opened 2021-09-21
Attendance Sort by ID Number

tribalHAT:
Attendance Sort by ID Number ('idnumer') field of moodle.
Credit goes to  [Babaso Aldar](https://moodle.org/user/view.php?id=2043935&course=5), for his [suggested code changes ](https://moodle.org/mod/forum/discuss.php?d=377645)


renderer.php
```diff
@@ -917,23 +917,36 @@ private function construct_fullname_head($data) {
...
+            $idnumber = html_writer::link($url, get_string('idnumber'));
         } else {
+	        $idnumber = html_writer::link($data->url(array('sort' => ATT_SORT_IDNUMBER)), get_string('idnumber'));
             $firstname = html_writer::link($data->url(array('sort' => ATT_SORT_FIRSTNAME)), get_string('firstname'));
             $lastname = html_writer::link($data->url(array('sort' => ATT_SORT_LASTNAME)), get_string('lastname'));
         }
 
-        if ($CFG->fullnamedisplay == 'lastname firstname') {
-            $fullnamehead = "$lastname / $firstname";
+        if ($CFG->fullnamedisplay == 'idnumber firstname lastname') {
```
danmarsden, line 944: this all feels a bit to specific to your use of idnumber - usually people would drop it in a new column rather than in the user fullname field - I think we should avoid these changes and just add the sort option.

but... I'm also hoping post 4.0 release we can re-do a chunk of the user reports using reportbuilder code which will hopefully give us some better control over columns/sorting etc.

tribalHAT: Yes, you are right and adding the sort option would be great.

danmarsden:
closing this one as the long term plan is to use some more of the reportbuilder api stuff.

PR #560, opened 2021-09-28
Redo SVG icon

lucaboesch:
I think the redo icon could use a SVG variant file.


danmarsden:
nice - I've hated that one for a while so good to have something that looks tidier!!!
btw... while you're putting together SVGs..... feel free to put something together for Moodle 4.0 that matches the work they've done in MDL-71457 :-)

PR #571, opened 2021-12-08
make tableheader sticky

### LINKED ISSUE #559: First row header sticky
Hi Dan,

Is it possible to make the first row of the table "sticky" where the statuses are on the taking attendance table? 

The reason is that we have 5 - 6 statuses and our courses and 30-40 students per course. Instructors would like to see the statuses while scrolling down when are taking attendance.

Thank you!

t-schroeder:
fixes #559

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -300,3 +300,36 @@
 .attendancereporttable img.icon {
     padding-left: 5px;
 }
+
+/* Make tableheader sticky. */
+#page-mod-attendance-manage th,
+#page-mod-attendance-report th,
+#page-mod-attendance-preferences th,
+#page-mod-attendance-view th {
+    position: sticky;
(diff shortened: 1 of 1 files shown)


danmarsden:
Thanks Tim - which themes have you tested this with so far? - I hope you've tested both Boost and classic (core themes) but it would be good to confirm that and know what else you have tested this with so far before I do.

danmarsden:
closing this one as it doesn't appear to work in Moodle 4.0's theme - I'm still open to receving a css-based fix for this, but longer term as mentioned in the other issues I'd like to replace some of the reports with Moodle's reportbuilder API and let that deal with rendering.
thanks!

t-schroeder:
I've created a new pull request for Moodle 4. See #647.

PR #591, opened 2022-05-25
Fix All Sessions report when there are hidden sessions.

### LINKED ISSUE #590: "all sessions" report breaks when some are not visible to user
All sessions report fails with an error when there are hidden sessions.

nwp90:
Fixes #590 in this branch; similar fix needed to MOODLE_400_STABLE branch.


danmarsden:
generated a patch for 400_STABLE and merged in manually - thanks for the PR!

PR #600, opened 2022-06-28
Add support for custom identity fields; fixes #548

### LINKED ISSUE #548: Follow-Up on Issue #546 - Custom Profile Field Ignored, But Title Displayed in Title Row
Thank you for resolving the problem with custom profile fields causing a failure in the Attendance Module.  Now, the Attendance Module allows you to record attendance (even with a custom profile field), but it lists the name of the field in square brackets at the top of the screen where you record attendance and at the top of the Reports screen as well.  That field title display is confusing to faculty users.  I wonder if there is a way to prevent that title from displaying.  
Attendance_custom_profile_field_displays_taking_attendance
Attendance_custom_profile_field_view_Report_screen

mackensen:
This change adds support for custom user profile fields:

- Ensures that the custom profile data is loaded on the report page and on the task page.
- Uses `\core_user\fields::get_display_name()` to ensure that identity field names are consistently rendered
- On the export page, excludes the custom field from extra fields since they're loaded later. This restored the expected behavior that custom fields have to be explicitly allowed for exporting by the admin. This does not affect the report page.

I can prepare a backport for 3.11.

### DIFF
--- a/classes/form/export.php
+++ b/classes/form/export.php
@@ -111,7 +111,7 @@ class export extends \moodleform {
             $checkedfields['ident[id]'] = true;
         }
 
-        $extrafields = \core_user\fields::for_identity($modcontext)->get_required_fields();
+        $extrafields = \core_user\fields::for_identity($modcontext, false)->get_required_fields();
         foreach ($extrafields as $field) {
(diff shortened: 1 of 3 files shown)


danmarsden:
thanks for the PR - looks good to me - merged it in.

PR #605, opened 2022-08-12
Fix divide by zero error when viewing course summary report

svanschuyver:
## Problem Description
When viewing the course summary report at the activity module level, the page errors if the points and maxpoints for the assignment activity are set to zero. SQL throws an error when dividing 0 by 0.
image

## Resolution
Use NULLIF function to divide by null if maxpoints is zero. This will prevent the error from occurring, and the page will load as expected.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -982,7 +982,7 @@ function attendance_course_users_points($courseids = array(), $orderby = '') {
         $params = array_merge($params, $inparams);
     }
 
-    $sql = "SELECT courseid, coursename, sum(points) / sum(maxpoints) as percentage FROM (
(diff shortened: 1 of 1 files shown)


danmarsden:
Thanks for the PR - I've had a look through the Moodle codebase and I can't see NULLIF() used anywhere in existing code which is making me reluctant to use it... but... it does look it's supported by postgres/mysql/oracle/ms sql so it might be ok - will try to come back to this and look at it a bit closer at some point.

svanschuyver:
Yeah, I've wondered if there's a more "moodle" way to fix that code, but I didn't find anything, thus NULLIF().

danmarsden:
I went for a slightly different approch - that report is supposed to show an indication of average attendances - we can't do that for attendances that have 0 maxpoints so just easier to exclude them from the report.

PR #608, opened 2022-08-31
Remove custom tabs and move to using core secondary nav.

danmarsden:
dropping this in a PR so I don't forget.

rewrites the custom tabs/navigation to use Moodle 4's new navigation stuff - tidies up the page a bit and saves some screen real-estate.


danmarsden:
still to fix...
new add session button I don't think is quite in the right place, and it should not show on the report page in the location it is currently showing via this patch.

PR #610, opened 2022-09-04
Add custom session fields

### LINKED ISSUE #502: Ability to add summary for session
It would be useful to be able to add an overall summary for the session, and for this to appear in reports. This would be useful for a teacher to record general progress of the class, etc. With that in mind, should only be shown to teacher.

### LINKED ISSUE #556: custom session fields
Hi, first of all, thanks for this module.
I would like to ask, if there is a possibility to add a few features in the near future.
I would need some more or some customized fields for the lesson event to fill out and extract, like: the following
:Place, Teacher,,     Duration (hours), Kind,      Course Type,       Course Mode
Rome, M. Draghi,  3,                          Update, Specific product, Presence

The last three could be dropdowns with a few values (like the status (present, excused, absent)

Also, the exported data should be with the course name in the row, not grouped by course
I only need records with confirmed presence, so I don't need the columns:
P	R	A	Lezioni registrate	Punteggio	Percentuale
Please see the attached example of the data I would need.
[PROVA REPORT IVASS.xlsx](https://github.com/danmarsden/moodle-mod_attendance/files/7138345/PROVA.REPORT.IVASS.xlsx)
Thanks for your kind attention,
Michael

danmarsden:
Fix #556 Fix #502

### DIFF
--- a/backup/moodle2/backup_attendance_stepslib.php
+++ b/backup/moodle2/backup_attendance_stepslib.php
@@ -60,6 +60,10 @@ class backup_attendance_activity_structure_step extends backup_activity_structur
             'preventsharediptime', 'caleventid', 'calendarevent', 'includeqrcode', 'automarkcmid',
             'studentsearlyopentime'));
 
+        $customfields = new backup_nested_element('customfields');
+        $customfield = new backup_nested_element('customfield', ['id'], [
+            'sessionid', 'shortname', 'type', 'value', 'valueformat']);
+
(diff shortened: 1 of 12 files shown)


danmarsden:
first version of custom profile fields.
Still to do/check:

- backup/restore
- update export process to include custom fields
- update import process to allow custom fields to be imported
- Update report-builder source to include custom fields.
- Update Web services to include custom profile fields.

PR #632, opened 2022-09-23
Fix #631, Only group members should be able to take attendance

### LINKED ISSUE #631: When group mode set to visible, self-selection shows wrong message
as mentioned here:
https://moodle.org/mod/forum/discuss.php?d=438479#p1765241

related to #543

danmarsden:
Even when visible groups is set, don't allow attendance to be recorded when the user is not a member.


danmarsden:
untested fix for #631.

PR #640, opened 2022-11-03
Update locallib.php

mcruzel:
The student code of "rotate QRcode" is too simple. People can cheat easily by sending the 4-digits student code by SMS.

I tested, it works.
Is it a good solution ?

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -1331,6 +1331,26 @@ function attendance_renderqrcode($session) {
 }
 
 /**
+* Generate student passwords for rotate QRcode
+*/
+
+function attendance_generate__rotateqrcode_studentpasswords($length=15) {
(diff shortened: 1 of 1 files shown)


danmarsden:
the qr code rotation was always intended as an MVP that I'd hoped to improve one day... but yes, I do think it makes sense to increase the complexity of that string - I've moved to using a 15 char random string using the core random_string() function and pushed that into all supported branches.

but - keep in mind, nothing is stopping a student doing a video-call with their friends, sharing the qr code via a video call and allowing the student on the other end to scan the qr code - it's also pretty trivial to get the url out of the qr code and send that on as well - rather than just copying the password, copying the qr code url and sending that on.

thanks for the suggestion! :-)

PR #643, opened 2022-11-09
SLIN1 mod_attendance: initial commit of responsive fixes.

stephensharpe:
**Responsive fixes to filter controls.**

* Migrated main filter controls from `<table>` to use flex properties
* Removed some spacing CSS to instead utilise bootstrap margin/padding classes
* Works for report tab filtering conditions, individual user view/modes and all sessions tab

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -82,8 +82,6 @@ class renderer extends plugin_renderer_base {
         $context->viewcontrols = $this->render_view_controls($fcontrols);
 
         return $this->render_from_template('attendance/filter_controls', $context);
-
-        return $o;
     }
 
     /**
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -60,48 +60,28 @@ class renderer extends plugin_renderer_base {
      * @return string html code
      */
     protected function render_filter_controls(filter_controls $fcontrols) {
-        $classes = 'attfiltercontrols';
-        $filtertable = new html_table();
-        $filtertable->attributes['class'] = ' ';
-        $filtertable->width = '100%';
-        $filtertable->align = array('left', 'center', 'right', 'right');
+        
+        $context = new stdClass();
 
         if (property_exists($fcontrols->pageparams, 'mode') &&
             $fcontrols->pageparams->mode === mod_attendance_view_page_params::MODE_ALL_SESSIONS) {
-            $classes .= ' float-right';
-
-            $row = array();
-            $row[] = '';
-            $row[] = '';
-            $row[] = '';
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks @stephensharpe - nice work on converting it to a mustache template too! - looks like a really good improvement!

one thing I don't like is the new placement of the "add session" button - in your new version it floats in the center of the row above the table - in the old version it just sits with the session groups drop list.... which I do like better than having it float in the center but... I wonder if we can find a better place for it...

Looking at the attendance page on it's own, I wondered about moving the add session button to appear to the right of the main  navigation tabs, but nothing else in Moodle presents like that. I also thought about moving the "add session" into a the top navigation tabs beside "settings" - but that feels wrong too when I look at how other plugins implement the navigation... so I think I'm back to just thinking we make it align beside the session group drop list rather than float in the center.... unless you can think of a better place to put it? :-)

stephensharpe:
Thanks for the feedback @danmarsden! Happy to move the markup around to give a better overall UX.

Interestingly, enough I see something different. Could you provide a screen of where the 'add session' button is appearing for you? Is this an admin or learner view you are seeing it?

image
image

Thanks!
Stephen

danmarsden:
looks like it's the extra group mode stuff missing in yours - try adding some groups to the course and setting it to use separate groups - see the screenshots below:
original view:
original

Your updated code:
newcode

stephensharpe:
Ah, of course @danmarsden. Thanks for the screens! 💪 

Yes if you have a look now, the 'add session' should now be placed to the left. I experimented a little on where else this could be placed. Without digging a little deeper into the potential of other table->template conversions, I've opted to keep it where it is. Familiarity was the main factor here (we don't want people getting confused of where they thought the add session existed previously).

I also opted for tighter spacing between the different elements (as the default column spacing is a little much), so in this case I opted for the `p-2` class and `m-n2` for the parent flex container.

I'd like to probably experiment with other elements and make them a little tidier and more responsive (especially considering the UX focus of 4.0). They'll come over at some point in the future via PRs for you to review.

Thanks!
Stephen

danmarsden:
thanks Stephen! - I've squashed that manually, fixed the broken coding guideline issues and behat tests and pushed it into the 4.0 and 4.1 branches here. while looking at this I also thought it was silly to show the "add session" button on the user level report - in particular when viewing the all sessions page so I also dropped in a commit to remove that button from the page when viewing details about a specific user.

danmarsden:
I haven't triggered a version bump yet, so this won't land in the plugins db until I make that change next which I'll probably do when the plugins db opens up the ability to flag 4.1 support in the next week or so.

PR #647, opened 2023-01-02
made header row(s) of tables sticky

### LINKED ISSUE #559: First row header sticky

t-schroeder:
fixes #559

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -272,8 +272,6 @@
 }
 
 #page-mod-attendance-report .attendancereporttable {
-    overflow-x: scroll;
-    overflow-y: visible;
     padding: 0;
 }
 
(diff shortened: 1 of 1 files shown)


t-schroeder:
This works with theme_boost, but not with theme_boost_classic. For it to work with the Classic theme you'd need to add 50px to the `top` property of the `th`s. But I don't know how to determine in CSS which theme is being used. Otherwise I could add rules specific to e.g. the Classic theme.

danmarsden:
I think not being sticky in classic is fine - as long as what we do here doesn't "break" the view in classic - I've tested this and it looks ok to me so I'm merging it in - thanks for the PR! :-)

PR #648, opened 2023-01-16
Add columns and filters for number of statuses per status acronym.

michaelkotlyar:


danmarsden:
Thanks @uzimike  - I think we can probably do some further tuning at some point to improve the performance of those queries - I suspect the joins will be slow as they are querying all attendance data for all courses, rather than the data that is in the report itself - but lets leave them for now and come back to them later once we do some further testing - merging this in now thanks!

PR #649, opened 2023-01-20
Session data columns

michaelkotlyar:


danmarsden:
Thanks Michael - we really need to put this into a Join statement rather than add to the existing base sql - some reports will not use these columns at all and so we don't want this massive sql to run on every report - it's fine to add the same join sql to multiple columns, but hopefully you can split this out into it's own join sql added into the columns themselves.

Can you please move this into it's own $sql param and then add it as a join in the columns that need to use it?

michaelkotlyar:
That makes sense. Sure thing!

sumaiyamannan:
@uzimike,
Thanks for separating the join for the new columns.

@danmarsden , 
There are behat failures on 401 so I have created a new PR on 402 https://github.com/danmarsden/moodle-mod_attendance/pull/665.

Regards,
Sumaiya

danmarsden:
merged in manually via a separate PR - thanks Mike!

PR #651, opened 2023-01-22
Feature development for self reporting absence.

sumaiyamannan:
Hi,

The includes the below:

-     Added checkbox to add availablebeforesession field in all the required pages.
-     If availablebeforesession is enabled in a status of a session then the global setting of studentscanmarksessiontime will be overridden for it.
-     Students can see and submit attendance for a status of a session which has availablebeforesession enabled.
-     availablebeforesession allow attendance only on future sessions and not past.

Testing instructions:

-     Create a course
-     Add activity 'attendance' to a course
-     Edit setting of the activity and go to 'Status set'.
-     Select 'Available Before Session' on any one or more statuses and save.
-     Add a session of a future date

Go to the course as a student and click attendance.
You will see the above selected status(es) shown here.

Regards,
Sumaiya

### DIFF
--- a/classes/form/studentattendance.php
+++ b/classes/form/studentattendance.php
@@ -48,7 +48,7 @@ class studentattendance extends \moodleform {
         // Check if user has access to all statuses.
         $disabledduetotime = false;
         foreach ($statuses as $status) {
-            if ($status->studentavailability === '0') {
+            if ($status->studentavailability === '0' && time() > $attforsession->sessdate) {
                 unset($statuses[$status->id]);
             }
             if (!empty($status->studentavailability) &&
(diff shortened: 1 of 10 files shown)


classes/form/studentattendance.php
```diff
@@ -56,6 +56,9 @@ public function definition() {
                 unset($statuses[$status->id]);
                 $disabledduetotime = true;
             }
+            if ($status->availablebeforesession == 0  && time() < $attforsession->sessdate) {
```
danmarsden, line 59: Thanks @sumaiyamannan  this looks good - just one thing on this line I think is slightly wrong:

the $status->studentavailability setting allows for status values to be open "before" the official start of the session - for example, the session may start at 1pm, but students may be able to start flagging their attendance as present a few minutes before the session start (if allowed by the teacher.)

so:
```
time() < $attforsession->sessdate
```
should be closer to:
```
time() < ($attforsession->sessdate + ($status->studentavailability * 60))
```

hopefully that makes sense?

danmarsden: heh - nope... doesn't make sense... $status->studentavailability isn't the var I was meaning needs to be checked there... its the session "studentsearlyopentime" I really meant needs to be checked - if that setting is set, we need to make sure that the status values are still available for student marking.

danmarsden:
thanks Sumaiya! - I've merged this into the main branch with a few further changes. - thanks heaps!

PR #655, opened 2023-01-30
fix mysql csv push

EsdrasCaleb:
this little code helps to upload csv in a MySQL database

### DIFF
--- a/classes/import/marksessions.php
+++ b/classes/import/marksessions.php
@@ -228,7 +228,7 @@ class marksessions {
             }
             $sesslog[$userid] = new stdClass();
             $sesslog[$userid]->studentid = $userid;
-            $sesslog[$userid]->statusset = json_encode($statuses); //to work on mysql
+            $sesslog[$userid]->statusset = implode(",",$statusmap);
             $sesslog[$userid]->remarks = '';
             $sesslog[$userid]->sessionid = $this->att->pageparams->sessionid;
             $sesslog[$userid]->timetaken = time();
--- a/classes/import/marksessions.php
+++ b/classes/import/marksessions.php
@@ -228,7 +228,7 @@ class marksessions {
             }
             $sesslog[$userid] = new stdClass();
             $sesslog[$userid]->studentid = $userid;
-            $sesslog[$userid]->statusset = implode(",",$statusmap);
+            $sesslog[$userid]->statusset = implode(", ",$statusmap);
             $sesslog[$userid]->remarks = '';
             $sesslog[$userid]->sessionid = $this->att->pageparams->sessionid;
             $sesslog[$userid]->timetaken = time();
--- a/classes/import/marksessions.php
+++ b/classes/import/marksessions.php
@@ -228,7 +228,7 @@ class marksessions {
             }
             $sesslog[$userid] = new stdClass();
             $sesslog[$userid]->studentid = $userid;
-            $sesslog[$userid]->statusset = $statuses;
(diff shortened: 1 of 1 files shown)


danmarsden:
Hi @EsdrasCaleb - thanks for the PR - the data in the statusset column is supposed to be a simple csv list, not json-encoded data - I suspect your csv file might be invalid and the text isn't being escaped properly - can you please supply an example csv file that you are trying to import so I can take a look and see if there is anything we can do in the code to make it more reliable for you?

thanks,

danmarsden:
I'm closing this off as it doesn't appear valid as I mention, if you are having issues uploading a csv using mysql, please report that as an issue and supply the csv file that you are having issues with so I can try to reproduce - thanks!

EsdrasCaleb:
I guess I was not clear. This fix is not for the csv. The MySQL database do not support this attribute as it is it expect a string not an object this is why I put the json_encode

EsdrasCaleb:
image
This is the error I get if I do not do the json_encode where I did

EsdrasCaleb:
the code just change the way it useses to serve the log not the data

danmarsden:
can you please supply an example csv file that I can use to reproduce this?

EsdrasCaleb:
Its not the CSV its the database if you upload this in MySQL database you will get this error
[chamaAula2.csv](https://github.com/danmarsden/moodle-mod_attendance/files/10551858/chamaAula2.csv)

EsdrasCaleb:
This will not work in you plugin because the status collun is dependent of the language. This one is for english
[chamaAula2.csv](https://github.com/danmarsden/moodle-mod_attendance/files/10551916/chamaAula2.csv)

danmarsden:
thanks - it just helps to see what you're doing when reproducing - I suspect the correct fix there is is to use implode - not json_encode

eg something like:
```
$sesslog[$userid]->statusset = implode($statuses, ',');
```

do you want to give that a go?

danmarsden:
.... it looks like $statuses is coming through as an array in that location, when in should be a csv - implode will generate the data in the format it's "supposed" to be rather than json_decode.

danmarsden:
whoops- wrong order of params in that implode - should be implode(',', $statuses)

EsdrasCaleb:
I got this
image

EsdrasCaleb:
this variable is correct? it came from the database it is all columns from a table:
image

EsdrasCaleb:
If I use the statusmap instead statusset it works:
`$sesslog[$userid]->statusset = implode(",",$statusmap);`

danmarsden:
yeah good catch - statusmap looks right to me.

EsdrasCaleb:
Ok I changed

danmarsden:
cool thanks! - I suspect Moodle Plugin CI tests might fail (they are running right now) - missing a space after the comma in the implode call 
```
implode(",",$statusmap);
```
should be:
```
implode(",", $statusmap);
```

feel free to fix up on your end so I can merge otherwise I'll try to get to this at some point!

thanks

EsdrasCaleb:
Right, I did it. If you could replicate this to the 4.0 too because one of my sites is locked at this version :/...

danmarsden:
thanks @EsdrasCaleb - I've merged this in manually and cherry picked into the older supported branches as well - I don't currently plan to push out a plugins db release right now but will probably do that in the next few weeks along with the other development that is currently ongoing here at Catalyst - thanks again! :-)

PR #659, opened 2023-02-01
Allow student to update their attendance.

sumaiyamannan:
Hi,

This feature would allow a student to update their attendance for a session - in the example where they previously reported an apology, but managed to make it into the class anyway.

Testing instructions
----------------
- Create a course and add activity 'attendance' it
- Create session with current date and time
- Go to Attendance -> More -> Status set
- Tick 'Allow student to update' on one or more statuses. 
- Login as a student and go to attendance and record attendance, selecting of the above statuses
- Then go to attendance and click 'Update attendance'

Regards,
Sumaiya


locallib.php
```diff
@@ -575,6 +583,20 @@ function attendance_random_string($length=6) {
     return $string;
 }
 
+/**
+ * Does this status have allowupdatestatus enabled.
```
danmarsden: can we please change this function to a "record_exists" style call - and just return a boolean true/false in the return so we can simply do:
```
if (check_allow_update_status($sessionid))
```

danmarsden: also - looks like a naming violation there that I missed on the other function too... - plugin functions should be prefixed with the plugin name "attendance" unless they are in a namespaced class - these should be more like:
```
function attendance_check_allow_update_status()
```

Can you also please fix the other function I've already merged - in this same PR is fine! - thanks!

sumaiyamannan: I have fixed this and did a force push to add it in the same commit. 

Please note I could not change the $DB->get_field_sql to $DB->record_exists because I need a join statement, I believe I cannot do that in the latter.
But I have updated the function call to be able to use the below as the field value of allowupdatestatus can hold 0 or 1 value. 
`if (check_allow_update_status($sessionid)) `

danmarsden: sorry - I should have specified "record_exists_sql" - yep squashed force push is fine on this too :-)

sumaiyamannan: I have fixed this. I had to do some more changes to make it work with it. Please check.

danmarsden:
Thanks Sumaiya - sorry for missing this earlier - but I had intended that this would be a setting at the Attendance "session" level rather than at the status set level - eg when setting up a session, the teacher would set the time date/if the student can self-mark etc, and then also "can the student update their status if they have previously submitted a status. can you store this new setting in the attendance_sessions and put it on the session editing form instead of the status settings page?

sumaiyamannan:
@danmarsden I have now added this setting at the session level. This is done, all checks passed.

Issue #658, opened 2023-01-31
report absence UI


danmarsden:
When the student can submit an absence report, on the overview page the "submit attendance" link appears - just like it does when the session is open. We should change this so that it displays the text "Report future absence" if the session is not open yet and the session has status values that are available before the session starts.

sumaiyamannan:
This is done, please check the attached PR.

PR #660, opened 2023-02-02
Use different lang to report absence in the UI

sumaiyamannan:
Hi,
This is will fix the issue report absence UI #658.

Regards,
Sumaiya

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -1385,7 +1385,11 @@ class renderer extends plugin_renderer_base {
                         // URL to the page that lets the student modify their attendance.
                         $url = new moodle_url('/mod/attendance/attendance.php',
                                 array('sessid' => $sess->id, 'sesskey' => sesskey()));
-                        $cell = new html_table_cell(html_writer::link($url, get_string('submitattendance', 'attendance')));
+                        if (attendance_session_open_for_students($sess)) {
+                            $cell = new html_table_cell(html_writer::link($url, get_string('submitattendance', 'attendance')));
+                        } else {
+                            $cell = new html_table_cell(html_writer::link($url, get_string('submitattendancefuture', 'attendance')));
+                        }
                     }
                     $cell->colspan = 3;
                     $row->cells[] = $cell;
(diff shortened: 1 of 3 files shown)


classes/output/renderer.php
```diff
@@ -1385,7 +1385,11 @@ private function construct_user_sessions_log(user_data $userdata) {
                         // URL to the page that lets the student modify their attendance.
                         $url = new moodle_url('/mod/attendance/attendance.php',
                                 array('sessid' => $sess->id, 'sesskey' => sesskey()));
-                        $cell = new html_table_cell(html_writer::link($url, get_string('submitattendance', 'attendance')));
+                        if (time() < $sess->sessdate) {
```
danmarsden: I feel like we might need an api function "is open for marking" - because some sessions will be open for marking "early" and so still need to show the normal string and it's a bit messy how the rest of the code does this inconsistently.

sumaiyamannan: I have fixed this and did a force push to add it in the same commit. Let me know if in the future you prefer the feedback work to go in a separate commit.

locallib.php
```diff
@@ -575,6 +575,26 @@ function attendance_random_string($length=6) {
     return $string;
 }
 
+/**
+ * This functions checks if this session is open for students.
+ *
+ * @param stdclass $sess the session record from attendance_sessions.
+ * @return boolean
+ */
+function attendance_session_open_for_students($sess) {
```
danmarsden, line 584: can we make this function return "true" if the session is open and false if it isn't?

also - I'd look at simplifying the checks a bit - maybe something like:
```
$sessionopens = empty($sess->studentsearlyopentime) ? $sess->sessdate : $sess->sessdate - $sess->studentsearlyopentime;
if (time() > $sessionopens) {
    return true;
} else {
    return false;
}
```

sumaiyamannan: @danmarsden  I have fixed this. 

I wanted to highlight the fact that this is dependent on studentscanmarksessiontime  setting which is enabled by default, it stops a user from submitting attendance outside the duration of a session. 
So in the example of a lecture starting at 9am and the teacher has set "studentsearlyopentime" to 5min, at 8:54am the link should say "submit future absence", at 8:55am it should show "submit attendance". This scenario will be possible if studentscanmarksessiontime is not checked as otherwise at 8:54am the link to submit will just not be visible.

PR #665, opened 2023-02-08
Attendance report - new grade columns

sumaiyamannan:
**This is a draft at the moment.**

Hi,

This is based on https://github.com/danmarsden/moodle-mod_attendance/pull/649 but for MOODLE_402 branch. 

New grade columns
The existing built-in reporting functionality in the attendance plugin provides information around the grades,
however this information is not exposed in the report builder report source. The following columns are added as available columns in the report builder source:

• Taken sessions
• Points over taken sessions
• Percentage over taken sessions
• Total number of sessions
• Points over all sessions
• Percentage over all sessions
• Maximum possible points
• Maximum possible percentage

Report builder filters for each of the above columns are also added.
The existing session date filter will allow for the user to filter the grade and status count columns further
(except the current/previous week columns).

Regards,
Sumaiya


danmarsden:
I'm not a big fan of the massive sql query used to generate this data, but it seems to work ok and we can look at re-visiting the performance if we run into issues at a later date.

PR #667, opened 2023-02-09
Hide suspended enrollments from roster by default

pennedav:
Implements a simple enrolment status filtering mechanism to hide suspended enrollments from the roster by default. Implements a simple UI dropdown to enable normal behavior.

Very likely should be enhanced via some sort of admin setting to adjust default behavior if desired.

This patch does not provide any Behat tests and applies only to the 3.11 branch. Merging to Moodle 4 branch will require more work.

The lack of this functionality was causing a lot of frustration for instructors at my institution since we suspend enrollments instead of unenrolling (for various reasons) so this is a quick-and-dirty implementation to work around this.

### DIFF
--- a/classes/structure.php
+++ b/classes/structure.php
@@ -898,6 +898,16 @@ class mod_attendance_structure {
             }
         }
 
+        // Apply filter for enrol status.
+        // We're only implementing a binary at the moment: if 1 is NOT passed,
+        // don't include suspended enrollments, otherwise we DO.
+        if ('1' != $this->pageparams->enrolstatus) {
+            // Filter suspended enrollments
+            $users = array_filter($users, function (object $user) {
+                return ($user->enrolmentstatus == 0);
+            });
+        }
+
         // Add the 'temporary' users to this list.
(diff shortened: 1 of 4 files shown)


danmarsden:
Thanks for the PR - this does sound like it would be useful for others as well - I'd prefer to have this covered by behat tests before I merge this in too - any chance you might be able to write something to cover the change and test it?

pennedav:
I'll certainly try. I've not used Behat before so it'll take me a bit to do probably. My time is rather limited at the moment. I was reviewing the existing tests and it looks like [this scenario](https://github.com/danmarsden/moodle-mod_attendance/blob/92bf9003f515f97401f542bd09155506bca4a8c8/tests/behat/report.feature#L40) would be a good starting point to copy.

danmarsden:
this needs to be bundled up with the work in #684 and include support for the mobile app so that the experience is the same on the app and on the web version.

I'm closing this PR as it's currently incomplete, but feel free to resubmit a new one if you get time to spend more time on it in future! - thanks.

Issue #664, opened 2023-02-07
Table overflow issue


zishanj:
We have issue with the table layout where it get overflows the width:

![image](https://user-images.githubusercontent.com/1451292/217286595-8dbea805-08cd-403c-93f9-8729321d8484.png)

https://github.com/danmarsden/moodle-mod_attendance/blob/08cfe2be515ab5ea459a5a99e5c5a1c90713da01/styles.css#L324 needs to be removed as it overrides the `overflow-x` of class `.table-responsive` from `auto` to `initial` effecting the table responsiveness.

danmarsden:
Hi there, what version of moodle are you using? Also please keep in mind we only test and support the core themes. If you have a problem with a 3rd party custom theme you may need to resolve this issue within your theme rather than the attendance code. 

Can you reproduce the same issue with the core boost theme?

zishanj:
Yes its boost theme on latest 4.1.1.

danmarsden:
thanks - I can see this now when I set a background image in the boost settings - unfortunately removing the overflow-x: initial breaks the sticky header on the table (see #559) - I'm not sure if I will have time to look into this, but feel free to suggest other fixes. pinging: @t-schroeder - have you seen this before too?

thanks!

zishanj:
Headers are already sticky with https://github.com/danmarsden/moodle-mod_attendance/blob/08cfe2be515ab5ea459a5a99e5c5a1c90713da01/styles.css#L296. We only have to include `overflow-y: auto` and limit the height of parent div (if its really needed) so it will make the vertical scrollbar appear.

zishanj:
Other option is to use javascript [extension](https://bootstrap-table.com/docs/extensions/sticky-header/) for bootstrap which will detect the page scroll and makes the header sticky when its about to hide above the container.

danmarsden:
Thanks - I'm not a fan of using JS to do this sort of thing and would prefer a css-based solution, but feel free to submit a pull request with the change for me to review!

thanks!

t-schroeder:
Admittedly I hadn't tested this with a different background color. Another fix would be to just make the table cells non-transparent. Grader report does the same and that's a Moodle core page:

![image](https://user-images.githubusercontent.com/24255259/218722994-daad336e-2ff7-4a1b-9eea-fd9fc1942f2c.png)

zishanj:
[...]

PR #669, opened 2023-02-15
made tables opaque

t-schroeder:


t-schroeder:
This fixes the issue of tables being transparent which is noticeable when the table extends beyond the screen width. This has been discussed in #664.

danmarsden:
Thanks @t-schroeder !

PR #674, opened 2023-02-22
Fix initial attendance not being recorded when updating attendance status is allowed

### LINKED ISSUE #673: Problems with new setting "Allow students to update own attendance"
#659 introduced the option on session creation to allow students to update their own attendance status after already having submitted one. I tested it and noticed some problems:

1. When creating a session the new setting "Allow students to update own attendance" appears twice when I check "Allow students to record own attendance".
2. With the new setting enabled (both in site administration and for a session) a submitted attendance status doesn't get saved anymore. Updating an existing status works, but inserting a new one doesn't.

t-schroeder:
fixes #673

### DIFF
--- a/classes/form/studentattendance.php
+++ b/classes/form/studentattendance.php
@@ -78,11 +78,12 @@ class studentattendance extends \moodleform {
                 // Check if an existing status is set, and show it.
                 $existingstatusid = $DB->get_field('attendance_log', 'statusid',
                     ['sessionid' => $attforsession->id, 'studentid' => $USER->id]);
-                $existingstatus = $attblock->get_statuses(false)[$existingstatusid];
-                if (!empty($existingstatus)) {
-                    $mform->addElement('static', '', '', get_string("userexistingstatus", 'mod_attendance', $existingstatus->description));
+                if ($existingstatusid) {
+                    $existingstatus = $attblock->get_statuses(false)[$existingstatusid];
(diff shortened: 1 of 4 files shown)


classes/form/studentattendance.php
```diff
@@ -78,8 +78,8 @@ public function definition() {
                 // Check if an existing status is set, and show it.
                 $existingstatusid = $DB->get_field('attendance_log', 'statusid',
                     ['sessionid' => $attforsession->id, 'studentid' => $USER->id]);
-                $existingstatus = $attblock->get_statuses(false)[$existingstatusid];
                 if (!empty($existingstatus)) {
+                    $existingstatus = $attblock->get_statuses(false)[$existingstatusid];
```
danmarsden, line 82: this doesn't look right - moving it here will mean the above check will always fail?

t-schroeder: Thanks, I just did that last commit in a hurry. It's fixed now.

danmarsden:
thanks - I didn't like the way that existing code was calling attendance_check_allow_update twice, so I've re-formatted that if statement a bit and pushed in manually. Thanks heaps for testing! - we're going through a bunch of user-testing on this at the moment with a client but it's hard to make sure everything is tested well... I'd hoped to get some more behat tests in but hopefully another day...

PR #675, opened 2023-02-22
fixed event data not containing attendance_log id

t-schroeder:
It looks like it was originally intended to work like this. Before with `false` being passed you'd only get back a boolean, usually `true` instead of the id of the database record.


danmarsden:
nice catch - merged that in manually too - thanks!

Issue #688, opened 2023-06-28
Course restore with mod_attendance prior to 2023020106 fails with newer version installed


ghost:
When restoring a course with an Attendance activity from a system with mod_attendance < 2023020106 to a system with that version or newer, the restore fails because a null value is supplied for allowupdatestatus. Suggest leaving the default 0 on this column but making it nullable. Will post a PR -- took a quick look at the code impacts and ended up just adding a COALESCE statement using the column default (0) to the SELECT statements.

ghost:
https://github.com/danmarsden/moodle-mod_attendance/pull/689

PR #689, opened 2023-06-28
Issue #688

ghost:


danmarsden:
thanks for the PR - sorry it took so long to come back to this - I've just pushed a slightly different fix for this in d8e07b860dfbaae4704bf00f604ad22dada5fe7e - I didn't have any old restore files around to test this with but it should resolve the issue reported - please let me know if you still have issues related to older backups though!

Issue #686, opened 2023-06-16
External tests not passing in Moodle 4.2


juancs:
The error we get is:

PHP Fatal error:  Uncaught coding_exception: Coding error detected, it must be fixed by a programmer: When including this file for a unit test, the test must be run in an isolated process. See the PHPUnit @runInSeparateProcess and @runTestsInSeparateProcesses annotations

This error is due to this change in Moodle 4.2:

https://moodledev.io/docs/4.2/devupdate

danmarsden:
Fixed in the merged PR - closing.

PR #690, opened 2023-06-29
Issue #686

aydevworks:


aspark21:
Fixes #686 
Alex also had a look at the draft from @juancs https://github.com/juancs/moodle-mod_attendance/commit/2cb14530b7701bbbc039774f1c63432ab5dec8a2

Believe this is the better fix

ties back to a core change in 4.2 https://tracker.moodle.org/browse/MDL-78049

PR #699, opened 2023-08-25
ATT-698 Fix deleting when enablecalendar is off

### LINKED ISSUE #698: Error Deleting Attendance When "enablecalendar" is False
## Steps to Reproduce

1. On a clean install, disable `enablecalendar` from the site plugin settings
2. Turn on debugging to 'DEVELOPER'
3. Create a course
4. Add an Attendance activity
5. Add multiple sessions - at least two
6. Delete the Attendance activity
7. View error in the output from the next run of admin/cli/cron.php

The error below occurs once for each session created:

Screenshot from 2023-08-25 09-16-56

This is cause by the DB call in classes/calendar_helpers.php:
https://github.com/danmarsden/moodle-mod_attendance/blob/cf4ffcf1ad73b5d358ddaa7c7bb8bf975d8dd55f/classes/calendar_helpers.php#L129

In attendance_sessions, `caleventid` is `0` for _all_ entries because `enablecalendar` is set to `false`. The call to `$DB->get_records_list()` returns an array of objects where the key is the value of the given field, in this case, `caleventid`. Since  all the `caleventid` values are `0`, this produces the error above.

I think one fix is pretty easy - I'll submit a PR soon that addresses this issue.

martygilbert:
This fixes #698  -  See that writeup for more information, but the essence of the issue is:

- Turn off `enablecalendar`
- Create an Attendance and add multiple sessions
- Delete the attendance
- Error shows in cron.php

This is because all `caleventid` values are zero if `enablecalendar` is false. An error occurs because instead of unique values for `caleventid` to construct the array in `attendance_existing_calendar_events_ids()`, you receive duplicate values for the keys (zero), so it complains.

## This Fix
This adds the `DISTINCT` keyword, which returns only one zero when `enablecalendar` is off, and all of the distinct entries when it's one. There are other ways to fix this, of course, but this seemed to introduce the least amount of change.

### DIFF
--- a/classes/calendar_helpers.php
+++ b/classes/calendar_helpers.php
@@ -175,7 +175,7 @@ function attendance_delete_calendar_events($sessionsids) {
  */
 function attendance_existing_calendar_events_ids($sessionsids) {
     global $DB;
-    $caleventsids = array_keys($DB->get_records_list('attendance_sessions', 'id', $sessionsids, '', 'caleventid'));
+    $caleventsids = array_keys($DB->get_records_list('attendance_sessions', 'id', $sessionsids, '', 'DISTINCT caleventid'));
     $existingcaleventsids = array_filter($caleventsids);
     if (! empty($existingcaleventsids)) {
         return $existingcaleventsids;


danmarsden:
Thanks for the PR and explanation - I've merged that in - will look at backporting to some of the older branches later!

PR #704, opened 2023-10-13
Performance increase on 2022090900 upgrade step.

jwcatau:
PostgreSQL performance increase for sites with large numbers of records in {attendance_log}.

Step 2022090900 upgrade on a database with 1.3 million records (3 duplicates) took 21-24+ hours and with the fix took 33 seconds.

Instead of a `DELETE ... WHERE NOT IN` this approach uses a `SELECT` to create a new "temporary" table, `TRUNCATE` the old one, re-`INSERT` the records back, and `DROP` the working table, in a delegated transaction.

```
# Total records in the table
moodle=# SELECT COUNT(*) FROM mdl_attendance_log;                                                             
  count  
---------
 1342090  
(1 row)

# Unique records not to be deleted from SQL sub-select:
moodle=# SELECT COUNT(1) FROM (SELECT max(id) FROM mdl_attendance_log GROUP BY sessionid, studentid, statusid) data;
  count                   
---------                                                                                                                                                                                                          
 1342087                                                                                                                                                                                                           
(1 row)                                


# Before SQL change:
-->mod_attendance
...
++ 2022090900: Success (75622.90 seconds) ++

# After SQL change:
-->mod_attendance
...
++ 2022090900: Success (33.51 seconds) ++
```

PR #710, opened 2023-10-25
Performance increase on 2021050700, 2022090900 upgrade steps.

petersistrom:
PR to update the fix raised in https://github.com/danmarsden/moodle-mod_attendance/pull/704

PostgreSQL performance increase for sites with large numbers of records in {attendance_log}.

Step 2022090900 upgrade on a database with 1.3 million records (3 duplicates) took 21-24+ hours and with the fix took ~30 seconds.


danmarsden:
seems abitrary but.. can we drop the use of "AS" in that query to make ci checks happy?  :-)

petersistrom:
No worries, I've removed the "AS"

PR #733, opened 2024-03-07
Use Server Time when rotating QR Codes

RedyAu:
This is my first contribution to Moodle, so please don't hesitate to correct me on any coding standard issues.

I have made a quick patch to rotating QR codes, so that the client's browser gets the server time, and rotates QR Codes using that. 
We've had repeated issues stemming from the teachers' machines not being in sync with our Moodle server, this PR fixes that: It doesn't matter what the client's clock is anymore.

I will submit this patch to Moode 3.X as well - that is what we still use and what I made first, but I ported the fix to the latest branch too (#734).

PR #734, opened 2024-03-07
Use Server Time when rotating QR Codes (Moodle 3.11.X)

RedyAu:
This is my first contribution to Moodle, so please don't hesitate to correct me on any coding standard issues.

I have made a quick patch to rotating QR codes, so that the client's browser gets the server time, and rotates QR Codes using that.
We've had repeated issues stemming from the teachers' machines not being in sync with our Moodle server, this PR fixes that: It doesn't matter what the client's clock is anymore.

I have submitted this patch to Moode 4.X as well - we still use 3.X and that is what I made first, but I ported the fix to the latest branch too (#733).

### DIFF
--- a/js/password/attendance_QRCodeRotate.js
+++ b/js/password/attendance_QRCodeRotate.js
@@ -15,11 +15,13 @@ class attendance_QRCodeRotate {
         this.qrCodeHTMLElement = "";
     }
 
-    start(sessionId, qrCodeHTMLElement, textPasswordHTMLElement, timerHTMLElement) {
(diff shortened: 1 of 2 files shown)


danmarsden:
Hi RedyAu - I'm just getting into some work on the attendance plugin again - I don't support branches for Moodle versions that are no longer supported by Moodle unless it's for something quite serious - support for Moodle 3.11 ended in December 2023 but hopefully as you can port this locally it won't really matter that I'm closing this one as won't fix - I will try to backport your patch to 4.1 and higher branches though. - thanks!

Issue #727, opened 2024-02-26
Include QR Code option misaligned when creating new session


andrewrowatt-masseyuni:
Hi Team,

When creating a new session, the _Include QR code_ option is adjacent to the _Random password_ option (screenshot 1). This is causing some confusion for some users who assume that the Password and QR code options are linked.

On the editing screen (screenshot 2) the _Include QR code_ option is positioned below as expected.

Screenshot 1
![image](https://github.com/danmarsden/moodle-mod_attendance/assets/5333879/70a279cf-dc0c-4153-b126-6e135ff43b73)

Screenshot 2
![image](https://github.com/danmarsden/moodle-mod_attendance/assets/5333879/50b27cdd-21e6-43fb-ba1e-b316e3f4939f)

danmarsden:
Thanks Andrew - feel free to raise an internal WR if you want us to look at this otherwise it may sit here in the backlog for a while.

thanks!

andrewrowatt-masseyuni:
> Thanks Andrew - feel free to raise an internal WR if you want us to look at this otherwise it may sit here in the backlog for a while.
> 
> thanks!

FYI WR429183 has been raised @danmarsden

danmarsden:
fix merged into main version - the team will look to update the version on your site with the fix soon.

PR #735, opened 2024-03-08
Fix position of Include QR code option

aneno-m-e:
Removed 'Include QR Code' form item from group to address #727


danmarsden:
Thanks @aneno-m-e - I've cherry-picked that into the 4.1 and higher branches.

PR #751, opened 2024-04-23
Type notice on installation

micaherne:
On installation with PHP 8.1 I'm seeing this message: `Notice: Object of class stdClass could not be converted to int in /var/www/html/mod/attendance/locallib.php on line 1258`

It looks like the $COURSE->id should be used for comparison with SITEID instead of the whole object.


danmarsden:
Arg - whoops. Late night coding tripped me up.there! Will merge it this morning, thanks for the PR!

PR #753, opened 2024-05-13
Remove special chars from acronym fieldnames #752

### LINKED ISSUE #752: Acronyms with special characters cause error when using Attendance as datasource in Report Builder
Hi Dan,

We ran into an issue where--when using Attendance as a data source in the report builder--if there exists a mod_attendance instance that has a status set that contains an acronym with a special character, this will cause a fatal error in the report builder. The error occurs because the Attendance plugin creates a column for every unique acronym in the system, using the acronym itself as part of the column name, but Report Builder requires that column names be alphanumeric; indeed, Moodle throws an error that says as much (I can't remember the exact error text at the moment).

This should be reproducible by creating a new Attendance module, going to the Status Set page, and adding a new row with an acronym like "A+". Then, go to the report builder and create a new report using Attendance as the datasource. You should encounter a fatal error that prevents the page from loading.

The issue can be fixed by either requiring that acronyms be alphanumeric, or by changing the code that generates the column names from the acronyms in the system to escape special characters. If you let me know which of these approaches you'd rather take, I can probably create a pull request. Let me know if you need any more information.

Limekiller:
This small change encodes the acronym field names in base64 (removing trailing equals) to fix https://github.com/danmarsden/moodle-mod_attendance/issues/752, where acronyms with special characters cause an issue because reportbuilder requires that filters and columns be alphanumeric. This keeps acronym names unique while removing special characters.

It does require removing the strtolower() call, since base64 uses both upper and lower case letters -- I hope that isn't an issue. Everything worked well when testing.

### DIFF
--- a/classes/local/entities/attendance.php
+++ b/classes/local/entities/attendance.php
@@ -739,7 +739,7 @@ class attendance extends base {
      * @return array
      */
     private function acronymfieldnames(string $acronym): array {
-        $fieldname = 'status_' . strtolower($acronym) . '_total_count';
+        $fieldname = 'status_' . rtrim(base64_encode($acronym), '=') . '_total_count';
         return [
             $fieldname,
             $fieldname . '_current_week',


danmarsden:
I vaugely recall field names needing to be lower case for some reason - but I can't remember why...... sounds like your testing would have identified if it's a problem though! - I'll try to review this properly sometime soon!

thanks!

danmarsden:
merged this in manually and cherry-picked to all supported branches - thanks heaps for the PR!

Issue #754, opened 2024-05-16
preconfigured Timeslots


Sythelux:
I'd like to have a preconfigured timeslot field for our teachers at school instead of the 2 drop-downs for the start time and end time each.

Also with current time being preselected. I attached a screenshot how I imagine it to be below.

![Bildschirmfoto vom 2024-05-16 21-31-45](https://github.com/danmarsden/moodle-mod_attendance/assets/601797/c7c1300d-ff01-4c2e-9ccd-9b3c6dee75be)

I can't pre bulk upload the attendance times. For reasons the time planning and moodle attendance are 2 separate systems at my school.

I think it is a lot to ask, but I still try.

danmarsden:
Hi @Sythelux thanks for the suggestion - there is definitely room for improvement around the session start and end field - we also have a request in #638 which mainly covers the ability to set a better sensible default value in the start and end time in the form.

One of the challenges is that your example above is quite specific and the majority of the users of the attendance activity want the flexibility to define a specific start and end time.

One day - I'd like to see a time selector a bit like the create event dialogue in Google Calendar - that allows suggested timings but also with the ability to type directly into the time fields to set a specific time.

I'm going to close this as won't fix as it's not something I plan to work on as a volunteer and the design suggested here decreases the overall flexibility of the plugin - I'm also not a fan of the UX suggested, but getting that right is a lot more work.

I am still open to receiving pull requests to improve the time picker - but they need to be built in a way that improves the general UX and don't make it harder for me to implement something like a google calendar time picker at some point in future.

thanks again!

Issue #638, opened 2022-10-19
Time field of new attendance session should be populated by default


Bwise1:
When adding new attendance session, as the Date field is being automatically populated by the Date of the Day, can the "From" Field of the Time to be populated by the current Time the session want to be added.
![moodle](https://user-images.githubusercontent.com/26597408/196618678-9d1dc743-c154-49db-a84d-599fb990d7af.png)

danmarsden:
I do find 00 : 00 to be a weird default value myself, but I'm not convinced it makes sense to set the default value to the current time - usually teachers would create the sessions prior to them being used.

I think it does make sense to tidy up this selector - it's also difficult for people to create "all day" sessions as mentioned in #464 

feel free to help working on this by proposing a code-based solution.

Bwise1:
the default value to the current time feature can be done only for add session, (don't know how easy it is to achieve, but I will take a look at the code base and see ). and if you are creating prior to the time, at least you are not starting from 00.00, you are starting from a time closer to the time you want to use the attendance. saves you more time.

danmarsden:
thanks - let me know any further thoughts you have - I'm still not convinced it makes sense to use the current time as the default - eg - it's 20:07 right now, and if I was creating a session as a teacher, I'm more likely to be wanting to create one for tomorrow - or later in the week during the day. - there's little point in using the minute value and setting that to 07 as it would be really unlikely that a session would be set to start at 7 minutes past the hour and it's unlikely a teacher would be creating a session for 20:00 in some organisations who operate during the 0900-1700 time frame.

we could drop in a static "default" value in the admin settings - allowing an admin user to pre-select a time/hour for both the start and end time that makes more sense. What might be cool is to have some form of default duration too - so that if someone changes the start time, it auto-matically adjusts the end time using javascript - so if the majority of the time an organisation has 1hr sessions, the teacher sets the start time to 09:00 and the javascript auto-adjusts the end time to 10:00.. but... what I'd really want to do is completely replace that session date start/end selector and make it operate in a much more user-friendly way... of course that's going to take a bit more investment of time though.

Sythelux:
> the default value to the current time feature can be done only for add session, (don't know how easy it is to achieve, but I will take a look at the code base and see ). and if you are creating prior to the time, at least you are not starting from 00.00, you are starting from a time closer to the time you want to use the attendance. saves you more time.

[...]

PR #755, opened 2024-05-25
Implemented default populated session

poetter-sebastian:
Implemented settings according to #754 and #638 

New settings are: 

- minutes steps e.g. 15=> 0,15,30,45 in the minutes selection box (get ignored if the populated elements doesn't fit)
- set current time to default session start
- set default session length
- set default time blocks for automatic allocation

grafik

grafik

grafik


poetter-sebastian:
I'v updated to code and removed the undocumented function

danmarsden:
Thanks for the PR - sorry for leaving this for so long - I don't really like the complexity that this patch adds - what I'd like to see is:
1) A new site-level setting to define the default "start" time.
2) A new site level setting to define the default "duration"

And then some javascript that automatically updates the "end" time when someone changes the "start time" by the default duration amount.

PR #756, opened 2024-06-03
Smaller monologo icon for Moodle ≥ 4.4.

lucaboesch:
Adopt icon size to 24×24 with a smaller content as other icons.


danmarsden:
Thanks for sending that through!

PR #757, opened 2024-06-06
Reset button

basileLeroy:
Adding reset buttons for individual sessions to each user. 
Resets will remove the user's session data from the table


danmarsden:
@basileLeroy thanks - it looks like you have a lot of custom changes on your codebase - not just the reset button - any chance you could please clean this PR up so that it only contains the suggested reset button? - I also notice that your use of learnerid as a parameter being passed to the scripts doens't do any security checking - eg checking if the learner being passed is the current logged in user or the user has admin capabilities and should be allowed to reset for any user?

basileLeroy:
Hey Dan, my appologies, this was not meant to be pushed to this repository, feel free to close the request.

Have a nice day,
Basile

danmarsden:
Thanks @basileLeroy - it does look like you have some interesting improvements on that custom branch - would be great to see some of those land in the main release - feel free to submit a PR in future if you get round to cleaning it up a bit!

PR #759, opened 2024-07-08
Codechecker fixes

lucaboesch:
Adhere to more recent Moodle Coding Guidelines.

### DIFF
--- a/classes/import/marksessions.php
+++ b/classes/import/marksessions.php
@@ -132,15 +132,15 @@ class marksessions {
     /**
      * Constructor - parses the raw text for sanity.
      *
-     * @param string $text The raw csv text.
      * @param mod_attendance_structure $att The current assignment
+     * @param string $text The raw csv text.
      * @param string $encoding The encoding of the csv file.
      * @param string $delimiter The specified delimiter for the file.
      * @param string $importid The id of the csv import.
      * @param array $mappingdata The mapping data from the import form.
      * @param bool $useprogressbar Whether progress bar should be displayed, to avoid html output on CLI.
      */
-    public function __construct($text = null, $att, $encoding = null, $delimiter = null, $importid = 0,
+    public function __construct($att, $text = null, $encoding = null, $delimiter = null, $importid = 0,
                                 $mappingdata = null, $useprogressbar = false) {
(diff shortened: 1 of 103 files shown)


classes/import/marksessions.php
```diff
@@ -132,15 +132,15 @@ protected function get_column_data($row, $index) {
...
      * @param mod_attendance_structure $att The current assignment
+     * @param string $text The raw csv text.
      * @param string $encoding The encoding of the csv file.
      * @param string $delimiter The specified delimiter for the file.
      * @param string $importid The id of the csv import.
      * @param array $mappingdata The mapping data from the import form.
      * @param bool $useprogressbar Whether progress bar should be displayed, to avoid html output on CLI.
      */
-    public function __construct($text = null, $att, $encoding = null, $delimiter = null, $importid = 0,
+    public function __construct($att, $text = null, $encoding = null, $delimiter = null, $importid = 0,
```
danmarsden, line 143: did you change the order of these params on purpose? - this doen'st seem to be a codechecker change?

import/marksessions.php
```diff
@@ -66,12 +66,12 @@
 
 $formparams = ['id' => $cm->id,
                'sessionid' => $pageparams->sessionid,
-               'grouptype' => $pageparams->grouptype];
+               'grouptype' => $pageparams->grouptype, ];
 $form = null;
 if (optional_param('needsconfirm', 0, PARAM_BOOL)) {
     $form = new \mod_attendance\form\import\marksessions($url->out(false), $formparams);
 } else if (optional_param('confirm', 0, PARAM_BOOL)) {
-    $importer = new \mod_attendance\import\marksessions(null, $att, null, null, $importid);
```
danmarsden, line 74: looks like here'e the matching change for the function parameter order change - I think we should probably do this as it's own commit rather than bundling into the "codechecker fixes" if there's a good reason we should do this.

lucaboesch:
Well spotted, @danmarsden.
I've added an individual commit for it.
Note that the catalyst automation seems still using actions/checkout@v3 instead of actions/checkout@v4.

Best,
Luca

danmarsden:
cool thanks! - merged it in.

PR #760, opened 2024-07-12
add course category to custom report

Ferenc-Frank-Lengyel:
Could you please add the course category entity to the Attendance custom report so users can add category info to their report? Most importantly filter courses by their categories as they often represent departments, colleges, and schools. This would be beneficial for everyone.

### DIFF
--- a/classes/reportbuilder/datasource/attendance.php
+++ b/classes/reportbuilder/datasource/attendance.php
@@ -19,7 +19,8 @@ declare(strict_types=1);
 namespace mod_attendance\reportbuilder\datasource;
 
 use core_reportbuilder\datasource;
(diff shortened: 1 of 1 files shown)


danmarsden:
good idea - thanks for the PR! :-)

Ferenc-Frank-Lengyel:
Do you have a date in mind when this will be available in the Moodle plugin directory, please? I wasn't completely self-less with the PR ;)

danmarsden:
I've just bumped the version in version.php file which has triggered an auto-build to moodle.org plugins db now.

In future - if you send through a PR including a version bump that should auto-deploy to plugins db on merge :-)

PR #761, opened 2024-07-22
Adapt mobile code to ionic 7

dpalou:
In this PR I'm removing the ionic3 code (the ionic5 app was released on 2021 so it should be safe to remove it).

I also created a new latest template to use the new Ionic syntax for inputs. The old syntax still works in the 4.4 version of the app (ionic 7), but it will stop working in the 4.5 version (ionic 8) so some inputs could be wrongly aligned without these changes. To make it easier to review the code, in the first commit I duplicate the template (no changes done in the HTML) and then I apply the changes in a new commit.

Finally, I also removed the usage of core-course-module-description component for newer app versions, it is deprecated and will be removed in a future version of the app. It should have been removed already, but we kept it in the app because some plugins still use it.


danmarsden:
Thanks Dani - I assume this is safe to backport to the moodle 4.3/4.2/4.1 branches here as well - if so, please let me know and I'll cherry-pick it across to the other branches.

dpalou:
Hi Dan! Yes, it should be safe to backport it, none of the changes I did are affected by the Moodle version.

danmarsden:
great thanks -  I've backported to 4.1, 4.2 and 4.3 branches. :-)

Issue #713, opened 2023-11-05
Display raw password and QR code at same time.


frs2022:
Dear All, 

We face issue, we have upgrade Moodle to 4.1 version, we notes the password in attendance activity dose not  shown and not rotate as OR code as shown in attached file:
![Passwrod not shown or rotate ](https://github.com/danmarsden/moodle-mod_attendance/assets/117255482/106f3ee5-b667-4405-9965-ccdcb14bf62c)

Before we upgrade the Moodle it was rotate and shown:
![Password rotate ](https://github.com/danmarsden/moodle-mod_attendance/assets/117255482/7a7b23d9-6c1a-467a-83b7-4daceff12476)

Can you please advise us 

Thank you for you support

danmarsden:
this was an intentional change - displaying the password on the same page is ugly and it's a bad user experience asking students to enter a long password that is hard to read within a short time frame - they should be using the QR code scanning process.

Ideally it would be good to understand "Why" you want the password to be displayed? - it's also been discussed here:
https://moodle.org/mod/forum/discuss.php?d=443512

frs2022:
Dear Danmarsden, 

Thank you for your reply and support. 

Some classes and lecture halls are too big and wide for some students to scan the code (students' device compatibility). Hence, the password effectively substitutes code scanning when their cameras fail to capture the code. In our context, the rotated passwords and QR codes are needed.

danmarsden:
thanks - happy for someone to add this back in a Pull request, but it needs to display nicely - not just tacked onto the bottom of the page.

zmartinovic:
I agree it would be nice for teachers to have an option to display password even when rotating QR code. It can be one more checkbox in add session options. This would also allow students to mark attendance using mobile app.

mdorferf:
Dear Danmarsden,
We would also be delighted if this function were to be implemented again. We suspect that some student devices cannot read the QR code correctly. Perhaps there is a connection with this issue: https://github.com/danmarsden/moodle-mod_attendance/issues/746
Thank you in advance for your support!

danmarsden:
@mdorferf - personally I think it would be better to analyse what's going on with the student devices you mention - identifying and fixing that would be more useful to the students than giving them a long jumbled password they have to enter on a mobile device keyboard.

If you have internal development capabilities, pull requests are always welcome. If you would like to fund my time to investigate/fix/improve the plugin feel free to reach out privately.

mdorferf:
[...]

PR #768, opened 2024-09-27
add a button to display text password with qrcode

gmanaud:
Hello,

Here is a proposal to fix [#713](https://github.com/danmarsden/moodle-mod_attendance/issues/713) (Display raw password and QR code at same time.).

I added a collapse button to show the text password below the QR code if needed.

Capture d’écran du 2024-09-27 09-45-30

### DIFF
--- a/js/password/attendance_QRCodeRotate.js
+++ b/js/password/attendance_QRCodeRotate.js
@@ -16,9 +16,10 @@ class attendance_QRCodeRotate {
         this.timeOffset = new Date();
     }
 
-    start(sessionId, qrCodeHTMLElement, timerHTMLElement, serverTime) {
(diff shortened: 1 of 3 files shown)


danmarsden:
Thanks @gmanaud - interesting idea having a button on the page to hide/show, but this page really just needs an overhaul to make it look better than just dropping the password at the bottom of the page.

The page just looks messy - it needs to look better visually - maybe centering the qr code, or putting a nice clear border around the page - something visually to make it look nicer and display better on the page.

danmarsden:
I'm closing this off as "won't fix" - I'm still happy to receive a pull request, but it needs to include an improvement of how this page displays at the same time.

PR #769, opened 2024-09-30
Turkish translation is added

mucan54:


danmarsden:
Hi There - thanks for the effort you've put into this! - only the english language pack is included in the github repository - all other translations use moodle.org's translation site here: https://lang.moodle.org/ - if you could please login to that site and submit your improvements to the turkish language that would be great!

PR #771, opened 2024-10-09
fix #770 add indexes on attendance_log for update 2022090900

### LINKED ISSUE #770: Upgrade on mysql get stuck on 2022090900
The upgrade step for the version 2022090900 takes forever on mysql with around 800000 rows in the table.

https://github.com/danmarsden/moodle-mod_attendance/blob/MOODLE_401_STABLE/db/upgrade.php#L745

gbarat87:
Adding the index makes the Sql goes faster, speciallly for mysql.


danmarsden:
closing this PR as won't fix - as mentioned on the linked tracker issue - happy for someone to make this a CLI script instead of including it in the upgrade.php if they want to send a PR along those lines, but this is too hacky for me to land like this.

PR #773, opened 2024-10-22
Load the attendance form in the same page and without caching

davidl37:
Load the attendance form in the same page and without caching so that the Back button takes the teacher back to the course page.


danmarsden:
Thanks @davidl37 I've just merged this in manually - it's a nice improvement! :-)

danmarsden:
- only on the 4.5 stable branch for now, but feel free to send through a PR with a patch for older branches if you want to.

Issue #772, opened 2024-10-15
Hidden and suspended courses still visible on the 'All courses' tab


Ferenc-Frank-Lengyel:
The same student is enrolled in at least 2 courses using the Attendance activity. The student has sessions recorded in both courses. And then, we have to hide one of the courses for whatever reason, so it is not available to any students. However, it still comes up on the 'All courses' tab if the student goes to the available course.
I have tried to hide the second course (that should be enough) and suspend the student's enrolment, but none of them helped.
The only solution was to hide the activity itself to remove it from the 'All courses' tab.

danmarsden:
Thanks for the report - If you have internal development capabilities, feel free to help further by submitting a patch for review.

PR #787, opened 2024-10-31
get active (visible) courses only

Ferenc-Frank-Lengyel:
this is a possible solution to resolve issue #772

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -207,7 +207,7 @@ function attendance_get_user_sessions_log_full($userid, $pageparams) {
 function attendance_get_user_courses_attendances($userid) {
     global $DB;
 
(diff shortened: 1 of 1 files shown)


danmarsden:
That looks like it may hide the tab for all users - and not just the student themselves?

I think there is a case where the teacher should still be able to see the report for courses that are hidden so we might still need to retain that access somehow.

I think we should check inside attendance_get_user_courses_attendances to see if the currently logged in user has the viewhiddencouress capabiliity and if so, get all users courses, and if not - get only the "active" courses.

danmarsden:
closing this as it's not quite right yet, but feel free to send a new pull request if you get it sorted!

PR #790, opened 2024-11-13
issue #789: claim support of Moodle 4.5

dmitriim:


dmitriim:
@danmarsden Looks like MOODLE_404_STABLE works for Moodle 4.5. Not sure if you'd like to spin up a new MOODLE_405_STABLE branch for consistency.

danmarsden:
Thanks @dmitriim - I decided to create a new branch in the end - particularly because 4.5 is an LTS release and I want to drop support for 4.4 a long time before October 2027 (when 4.5 support upstream is dropped.)

PR #795, opened 2024-12-03
Update structure.php

amalkowskiint:
why didn't you add sessionid to event


danmarsden:
Looks like a sensible thing to do - I've merged that into the latest branch for 5.0 and higher - thanks for the PR!

PR #802, opened 2025-01-17
Remove hardcoded prefix

ak4t0sh:
Recently I had to upgrade an old version of this plugin and faced issues (upgrade crash) due to this hardcoded prefix.


danmarsden:
thanks for the PR! - I've merged it in.

PR #809, opened 2025-02-03
Fixes #808. Allow taking of attendance for non-participation groups.

### LINKED ISSUE #808: Unable to take attendance for groups with participation flag not set
If a group has the participation field set to zero, it is not possible for a user to take attendance for it, even if they have the `moodle/site:accessallgroups` capability (or even site admin). This seems to be a new feature that was [added in Moodle 4.2](https://github.com/moodle/moodle/commit/958da5b67ec00bf1c716515ef7416286f41f9899), alongside group visibility settings, and shows as "Show group in dropdown menu for activities in group mode". It is forced off unless the group visibility is GROUPS_VISIBILITY_ALL or GROUPS_VISIBILITY_MEMBERS.

In particular, this means that staff can't take attendance for groups where the membership is hidden to members.

Image

To reproduce:

1. Create a group and set visibility to hidden.
2. Create an attendance activity and a session including this group
3. Go to take.php and attempt to take attendance

You will see an error like this:

Image

micaherne:
This patch fixes issue #808 by allowing users with the "moodle/site:accessallgroups" capability on the attendance activity to take attendance for any group.

### DIFF
--- a/take.php
+++ b/take.php
@@ -54,7 +54,8 @@ $pageparams->init($course->id);
 $att = new mod_attendance_structure($att, $cm, $course, $PAGE->context, $pageparams);
 
 $allowedgroups = groups_get_activity_allowed_groups($cm);
-if (!empty($pageparams->grouptype) && !array_key_exists($pageparams->grouptype, $allowedgroups)) {
+$accessallgroups = has_capability('moodle/site:accessallgroups', $context);
+if (!$accessallgroups && !empty($pageparams->grouptype) && !array_key_exists($pageparams->grouptype, $allowedgroups)) {
      $group = groups_get_group($pageparams->grouptype);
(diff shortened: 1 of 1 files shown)


danmarsden:
This looks like it helps, but I'm confused by the existing code checking:
array_key_exists($pageparams->grouptype, $allowedgroups)

isn't grouptype the type of group - not the groupid which we'd expect to find in $allowedgroups?

(I haven't dug into the code completely to trace that yet...

micaherne:
No it's the group id that's passed from the calling page. I was puzzled by that myself!

danmarsden:
wow that's ugly..... but it should prevent us fixing this with the bad naming so I've merged this in.

danmarsden:
must be friday night - that meant to say "shouldn't prevent that landing... :-) - thanks for the PR!!

PR #842, opened 2025-08-10
fix:  Update externallib.php to handle new classes in moodle 4.5

hamzakhalem:
to fix: 
Coding error detected, it must be fixed by a programmer: mod_attendance_external::add_attendance_parameters did not return a valid external_function_parameters object Coding error detected, it must be fixed by a programmer: mod_attendance_external::add_session_parameters did not return a valid external_function_parameters object Coding error detected, it must be fixed by a programmer: mod_attendance_external::get_courses_with_today_sessions_parameters did not return a valid external_function_parameters object

### DIFF
--- a/externallib.php
+++ b/externallib.php
@@ -33,6 +33,12 @@ require_once(dirname(__FILE__).'/classes/attendance_webservices_handler.php');
  * @copyright  2015 Caio Bressan Doneda
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
+use core_external\external_api;
+use core_external\external_function_parameters;
+use core_external\external_multiple_structure;
+use core_external\external_single_structure;
+use core_external\external_value;
+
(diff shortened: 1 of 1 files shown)


danmarsden:
this seems a little strange to me - the externalib is covered by unit tests which I would have expected to pick up this failure... also if we do need to add this to the class, we'll likely need to add it to the MOODLE_500_STABLE branch first and then backport to 405_STABLE if needed.

danmarsden:
rebased this in #864 - this class should really be migrated to a namespaced class in the classes folder - there are core Alias's defined for these classes which is why it wasn't picked up in the tests but it's not clear when those aliases will actually get full removed - although I'm still not clear why you received coding error messages because I couldn't reproduce that... in any case - it makes sense to land and use the real class names instead of the to be deprecated core aliases for these classes so I'll merge this in once tests pass on #864 - thanks!

PR #844, opened 2025-09-02
Randomize password on restore

emmarichardson:
This pull request adds a setting to allow passwords to be randomized when you copy or restore a class with attendance sessions.  It is a site wide setting configurable in plugin settings.  For those that use a lot of course templates that automatically generate attendance sessions, it removes the need to manually change the password and increases security.  Code for the password generation in restore activity class.


danmarsden:
Thanks @emmarichardson - I think if we're going to add a new setting to control the length of the password length that should apply to all cases where attendance_random_string is used - not just in the backup/restore?

I'd be tempted to remove the parameter from the attendance_random_string() function as I can't see it being used anywhere and instead just do a check inside the attendance_random_string() functino for the new setting and use that length.

emmarichardson:
I removed the length setting - wasn't really needed anyway.  Fixed the line indents and language file.

emmarichardson:
Sorry Dan - not sure where that colon disappeared to.  Bear with me - kind of new to this!

danmarsden:
no worries! - looks good to me, thanks for the PR!

PR #854, opened 2025-11-08
Fix stored XSS in attendance remarks affecting teacher & student data

functionaldude:
## Summary

This PR fixes a stored Cross-Site Scripting (XSS) vulnerability in the Attendance module where user-controlled remarks are rendered without proper escaping in multiple report views.

Affected code paths:
- `renderhelpers.php` – `construct_remarks_cell()` rendered raw `$text` inside HTML.
- `classes/output/renderer.php` – several occurrences where `$ucdata['remarks']` and `$sess->remarks` are directly inserted into table cells.

Since remarks can be entered by users with limited privileges but are later viewed by teachers and potentially administrators, this is a classic stored XSS scenario.

These changes follow Moodle’s own XSS guidance (escape on output using s() / context-appropriate helpers) ([source](https://moodledev.io/general/development/policies/security/crosssite-scripting#escaping-output)).

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -1363,7 +1363,7 @@ class renderer extends plugin_renderer_base {
                 $row->cells[] = $status->description.$updatelink;
                 $row->cells[] = format_float($status->grade, 1, true, true) . ' / ' .
                                     format_float($statussetmaxpoints[$status->setnumber], 1, true, true);
-                $row->cells[] = $sess->remarks;
+                $row->cells[] = s($sess->remarks);
             } else if (($sess->sessdate + $sess->duration) < $userdata->user->enrolmentstart) {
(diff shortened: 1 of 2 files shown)


danmarsden:
Thanks for the PR - teachers are the only users that should be able to add remarks but the capabilities don't flag this as an XSS risk and it completely makes sense to strip out tags from that field. Can you please just remove the comments that you've added to the lines (they violate coding guidelines without fullstops) but we also don't want to add a comment like that every time we use s().

danmarsden:
Hi @functionaldude - I've rebased those commits manually myself, including the removal of the superfluous php comments and cherry-picked into all supported branches. I'll look to push out a release in the moodle.org plugins db later in the week.

Issue #866, opened 2025-12-21
Support Course overview


lucaboesch:
Attendance should support course overview

@see
https://moodledev.io/docs/5.0/apis/plugintypes/mod/courseoverview

pkgh:
Hi,
I have added an Attendance instance to my course, but it fails to show in the Attendances list on the new Course overview / Activities sub-page. All I get is the spinner trying to load something interminably.

Using Moodle 5.0.4+ (released 19 December, so the latest stable), and the latest Attendance (v. 2025122100) downloaded from the Moodle plugins page. I'm using the Boost Union theme, but I have also tested the plain Boost - to the same effect.

Been testing several other course module plugins, and so far only this one fails to show anything.

Attached is my screenshot - my site has the dev debugger turned on, but there's no warning or error.

<img width="1658" height="839" alt="Image" src="https://github.com/user-attachments/assets/2a047f3a-64e1-4d93-ac90-28c581e636a9" />



By comparison, on my MoodleCloud course page that's using Boost + Moodle  Version 5.0.2 (Build: 20250811) plus Attendance version 2025080800, the overview page does load the basic info at least - copying the plain text messages below
Attendance information not available.
Go to the Attendance overview.
plus the table with the link to attendance module
Table listing all Attendance activities Name
Attendance (2026-01-02 17:03)


General

The Attendance overview page ../mod/attendance/index.php?id=5  is also working OK there

When I try a link like this on my site, I get redirected to
../course/overview.php?id=23&expand[]=attendance#attendance_overview_collapsible

and these warning / errors (not sure all of them are legit):

Exception - Undefined constant core\output\local\properties\button::BODY_OUTLINE

[More information about this error](https://docs.moodle.org/500/en/error/moodle/generalexceptionmessage)
Debug info:
Error code: generalexceptionmessage
Stack trace:

    line 58 of /mod/attendance/classes/courseformat/overview.php: Error thrown
    line 250 of /course/format/classes/output/local/overview/overviewtable.php: call to mod_attendance\courseformat\overview->get_actions_overview()
    line 143 of /course/format/classes/output/local/overview/overviewtable.php: call to core_courseformat\output\local\overview\overviewtable->load_overview_items_from_activity()
    line 58 of /course/format/classes/output/local/overview/overviewtable.php: call to core_courseformat\output\local\overview\overviewtable->load_all_overviews_from_each_activity()
[...]

PR #870, opened 2025-12-23
Course activities overview

lucaboesch:
I just realized this could stumble with Moodle 5.0.

### DIFF
--- a/classes/courseformat/overview.php
+++ b/classes/courseformat/overview.php
@@ -52,10 +52,18 @@ class overview extends \core_courseformat\activityoverviewbase {
         );
 
         $text = get_string('view');
+
+        if (defined('button::BODY_OUTLINE')) {
+            $bodyoutline = button::BODY_OUTLINE;
+            $buttonclass = $bodyoutline->classes();
+        } else {
+            $buttonclass = "btn btn-outline-secondary";
+        }
+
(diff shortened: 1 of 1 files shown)


danmarsden:
I  prefer to just have a different branch to support different versions rather than adding code like this which really just introduces technical debt.

lucaboesch:
Is there a MOODLE_501_STABLE branch or are you working on it?
The 
`button::BODY_OUTLINE`
only works on Moodle 5.1, previous to that this didn't exist.
So at the moment as is Attandance on Moodle 5.0.4 would just silently fail when clicked in Course activity overview.
I unfortunately looked at it only closer in Moodle 5.1.1.

danmarsden:
If you get a moment just create a pr for 5.0 to switch to btn-outline-secondary  and i'll create a new 5.1 based on the existing branch. Thanks!

lucaboesch:
This comment https://github.com/danmarsden/moodle-mod_attendance/issues/866#issuecomment-3705717995 refers to the issue here.

danmarsden:
forgot to close this one as we fixed this on the 500_STABLE branch and left button::BODY_OUTLINE
 in place on the new 501_STABLE (but let me know if I missed something!!

Issue #851, opened 2025-11-03
Form Improvements: "Select activity" shown then hidden after a few seconds is confusing


tlock:
We would like to report the issue that in a big enough course "Select activity" is shown for seconds on adding or updating a session which is confusing for teachers when "auto marking" is disabled.

Suggestion is changing:

            $mform->addElement('select', 'automarkcmid', get_string('selectactivity', 'attendance'), $automarkcmoptions2);                                                                                                                                                                                                                                                                                                    
            $mform->setType('automarkcmid', PARAM_INT);
            $mform->hideif('automarkcmid', 'automark', 'neq', '3');
            if (!empty($sess->automarkcompleted)) {
                $mform->hardFreeze('automarkcmid,automark,studentscanmark,allowupdatestatus');
            }

To not be visible based on configuration instead of leaving mforms to handle it.

danmarsden:
makes sense thanks @tlock - feel free to help further by submitting a PR with the change.

PR #873, opened 2026-01-05
Fix the Select activity shown for seconds on adding or updating a session. #851

alexarevalo-l:
The fix initially hides the Select activity option when the page loads. Once the page has fully loaded, the JavaScript takes over and applies the hideIf logic, showing or hiding the option as appropriate.


danmarsden:
this looks like it hides the setting completely on the add session page - you have to save the session with the automark based on completion state - then go back and edit the session again to set it.

It does look like something is causing a performance impact on that page though - it shouldn't be taking that long to render and then trigger the client side JS so definitely an issue there - I've created #873 related to this, but this patch doesn't look quite right as the solution for now - thanks for the report.

If you do work out what's causing the performance impact or if you find a way to clean it up that doesn't mean you have to create a session - then edit it again to set a coursemodule feel free to send through another PR!

PR #874, opened 2026-01-06
added course wise support for students marking attendance after session time

zishanj:
By default, there is a single global check in plugin settings to allow students to mark attendance within session time. With this feature, there is a course level support to allow students to mark attendance after session for specific course even there is a global option enabled which restrict students marking within session time.

### DIFF
--- a/classes/structure.php
+++ b/classes/structure.php
@@ -75,6 +75,9 @@ class mod_attendance_structure {
     /** @var string subnets (IP range) for student self selection. */
     public $subnet;
 
+    /** @var bool flag set when students can mark after session time. */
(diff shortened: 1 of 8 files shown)


danmarsden:
thanks for the pr! Looks like this might be missing backup api support, import and web services support, as this is also a new feature it would be good to get some automated tests added to cover it before we merge this in.

danmarsden:
I've created #886 to track this as a feature request along with a link to this PR but for now as the work is incomplete I'm closing this off as an active PR. 

If you do get round to adjusting it as mentioned about feel free to send an updated PR through for me to take another look!

thanks again!

PR #882, opened 2026-02-03
Fix file retrieval condition in lib.php

ZhangTao08:
the multimedia inserted into description, they cannot be displayed with error "The media could not be loaded, either because the server or network failed or because the format is not supported."


danmarsden:
thanks for the PR - looks like it makes sense -I've pulled that in manually to the 5.1 and 5.0 branches here in github - thanks again!

Issue #887, opened 2026-02-16
"Output can not be buffered" error when dowloading Course Summary report on moodle 4.5


Wallywon:
When trying to dowload the "course summary" report from the attendance settings page, a conding error occurs : "Output can not be buffered before instantiating table_dateformat_export_format". 

Environment : 
Moodle 4.5.4+ and 4.5.5+
Plugin version : 2024082400

Steps to reproduce : 
1. Go to Administration > Plugins > Attendance > Course summary 
2. Click on the dowload button 
3. The error appears immediately

The error is triggererd at line 64 of coursesummary.php because some output has already started before the table export is initialized.

danmarsden:
as mentioned on #888  - I don't think this is an Attendance code issue - I suspect something in your stack is printing debugging message output or output to the screen before it's supposed to.

Can you track down what is being output to the screen? - it's likely to be a problem for other core code as well as this file download in the attendance plugin.

danmarsden:
sorry - I wasn't clear there, I have tried to reproduce this and haven't been able to.

Wallywon:
Thanks for you're investigation. 
I tried again on 2 plateforms ( A clean build on 4.5.2 and a sandbox on 4.5.7 ) and i could not reproduce the issue with the old version of the plugin. 
I tested with both my custom theme and the Boost theme but found no problems.

It's possible there is a conflict between two plugins. If I have time, i will investigate further and get back to you.

Here is the error I encountered in debug mode when i managed to reproduce the issus : 

```
Erreur de programmation détectée. Ceci doit être corrigé par un programmeur : Output can not be buffered before instantiating table_dataformat_export_format

Plus d’informations sur cette erreur

Info de débogage 
Error code: codingerror×Ignorer cette notification
Trace de la pile 
line 57 of /lib/table/classes/dataformat_export_format.php: core\exception\coding_exception thrown
line 214 of /lib/table/classes/flexible_table.php: call to core_table\dataformat_export_format->__construct()
line 198 of /lib/table/classes/flexible_table.php: call to core_table\flexible_table->export_class_instance()
line 64 of /mod/attendance/coursesummary.php: call to core_table\flexible_table->is_downloading()
×Ignorer cette notification
Tampon de sortie  ﻿
```

danmarsden:
[...]

PR #888, opened 2026-02-16
Fix export filename and output buffer for downloads

Wallywon:
This PR fixes 2 issues related to the course summary report export : 
1. Coding error : fixes "output can not be buffered before instantiating table_dataformat_export_format" occuring on Moodle 4.5 ( issue #887 ) 
2. Filename Cleanup : Fixes the double extension issue ( e.g. attendancecoursesummary.csv.csv ) by removing the hardcoded extension from the export filename

### DIFF
--- a/coursesummary.php
+++ b/coursesummary.php
@@ -51,8 +51,8 @@ if (empty($category)) {
 }
 // Check permissions.
 require_capability('mod/attendance:viewsummaryreports', $context);
-
-$exportfilename = 'attendancecoursesummary.csv';
+//fix : suppression of the .csv at the end because it is added automatically so when dowload we have a file attendancecoursesummary.csv.csv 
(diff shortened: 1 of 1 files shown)


danmarsden:
I've just manually added the fix for the double csv filename issue on all supported branches (NOTE: all fixes must land in the latest branch before being backported to older branches)

There shouldn't be any output being generated by the attendance plugin page prior to that call so if something is outputting data we need to fix that - rather than adding the ob_clean() call.

This sort of thing used to happen a lot when things like lang string files had a closing php tag with a carriage return - I wonder if something in your stack/theme or plugins adds extra output to the screen before moodle's standard $output->header() is called.

but - I'll do a quick check to see if I can reproduce the issue you have and comment on #887 with my findings.

PR #899, opened 2026-04-21
Fix access control in update_user_status

LeviG3:
The `update_user_status external` function let any authenticated user with `mod/attendance:view` mark attendance for another student, as long as the session allowed student self-marking. 
The capability check only gated teacher-style access; it didn't verify that a self-marking student was acting on their own record.

Changes:
externallib.php: require non-teacher callers to use their own studentid and takenbyid before allowing self-marking.

external_test.php: add a regression test that a student cannot mark another student via the web service.

### DIFF
--- a/externallib.php
+++ b/externallib.php
@@ -549,6 +549,11 @@ class mod_attendance_external extends external_api {
 
         // If not a teacher, make sure session is open for self-marking.
         if (!has_capability('mod/attendance:takeattendances', $context)) {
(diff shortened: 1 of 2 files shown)


danmarsden:
thanks heaps for catching that and sending through a patch! - I've pulled that through manually to all supported branches with a small change to move the global $USER call to the top of the function - will push out a new release to the plugins db shortly.

Issue #857, opened 2025-11-17
Course/session information on QR Code


mgmodell:
Can we get course information added to the QR Code scanning page? I teach courses back-to-back and tend to leave QR codes up through the entire class session. Then, when my next class walks in, they scan the old code while I'm busy in conversation. Having some identifying information on screen would help resolve this.

This seems related to #817 and maybe even #713 (my students, too, run into the password issue due to network issues in the room).

mgmodell:
@danmarsden , it looks like I could make this work by adding something like:
`.COURSE->shortname` into an `echo html_writer::div` on or around ln 65 of `password.php`, right?

My bigger question is: are there any tests of this QR functionality? I was looking in both `behat` and `external` and couldn't identify anything that looked promising. Is there anything? Did I miss it?

danmarsden:
Hi @mgmodell, yes I agree - that sounds like a sensible thing to have added to the page - unfortuantely you are correct, there aren't any automated tests that cover the QR handling code but it's an area I'd love to rewrite at some point and that would also include the addition of automated tests.

mgmodell:
@danmarsden, can you offer pointers towards setting up a dev. environment that I might use to build and test this? I'd like to help out but it's been a long time since I've done much in moodle or even in php. I'd rather not reinvent the wheel.

danmarsden:
Try the moodle academy courses  eg: https://moodle.academy/enrol/index.php?id=29

PR #900, opened 2026-04-22
Adding courseinfo to QR Code screen and devcontainer build setup

mgmodell:
This should satisfy #857 .
Full disclosure, I used copilot to help as I was struggling to get the dev environment set up. Therefore, in addition to the small addition to the QR Code screen, we also have. devcontainer setup that's been debugged and verified to work on my M1 mac using vscode. If you have the extension installed, it will build the moodle environment and launch it upon 'reopen in dev container.' There's also a script in the `/scripts` folder to create a course with a few students and an attendance. That way I was able to launch and test everything pretty quickly.

I fiddled a bit with the layout to find something that I like, but it might break some style rules and it's easy enough to change if needed. I suspect most people won't like it as is. Happy to help if so, but this is at least functional.

Finally, not sure the policy on version bumps, but I tried to update it appropriately.

### DIFF
--- a/classes/calendar_helpers.php
+++ b/classes/calendar_helpers.php
@@ -24,7 +24,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-require_once(dirname(__FILE__) . '/../../../calendar/lib.php');
+require_once($CFG->dirroot . '/calendar/lib.php');
+require_once(dirname(__FILE__) . '/../lib.php');
 
 /**
  * Create single calendar event bases on session data.
--- a/password.php
+++ b/password.php
@@ -56,6 +56,15 @@ $PAGE->set_title(get_string('password', 'attendance'));
 
 echo $OUTPUT->header();
 
+// Display course and session information to help teachers identify the active session.
+$course = get_course($cm->course);
+echo html_writer::tag('h2', 
+    html_writer::tag( 'strong', format_string($course->fullname, true, array('context' => $context)))
+);
+echo html_writer::tag('h3', 
+    html_writer::tag( 'strong', userdate($session->sessdate))
+);
+
 $showpassword = (isset($session->studentpassword) && strlen($session->studentpassword) > 0);
 $showqr = (isset($session->includeqrcode) && $session->includeqrcode == 1);
 $rotateqr = (isset($session->rotateqrcode) && $session->rotateqrcode == 1);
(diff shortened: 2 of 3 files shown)


danmarsden:
Thanks @mgmodell - best to pull out all the devcontainer stuff from the PR if you would like me to consider this - I don't think I've seen that in any moodle plugins and it's not something I'm currently looking at doing.

Also - I'm not quite sure if simply adding header tags is enough to make it display well - might be worth you dropping in a screenshot of how it looks into the PR as well?

Once that's done, please squash your commits into a single commit - I'm also a strong advocate for people using git commit trailiers when their code has been created with the use of AI tooling. - for example adding something like the following to the end of your commit message:
Ai-assisted-by: Copilot

thanks!

mgmodell:
Understood @danmarsden .

Here's what the current revisions look like. I'm happy to make changes, just let me know what you'd prefer. The page is already pretty simple and this gives me what I needed from it, but I was aiming for functional since I wasn't sure where to look for a design/style guide. Is there one I ought to consult?

I generally do not show the password as I use the QR with it embedded:
QR - no password

This is what it looks like with the password on screen:
QR with password

danmarsden:
thanks for adding the screenshot - I think it might look better to remove the "course:" and "Session:" text - just display the course fullname, and then underneath that show the session date. (no need for the session date to show on both lines either) - but other than that - using the header tags looks sufficient for now.

mgmodell:
Gotcha!
So, actually, the script threw the date onto the attendance name to make it distinct - it wasn't built in or anything. Below is a revised version which gets rid of the labels and also uses a revised name without the date. Is this good? Shall we go with it?
Revised QR (no password)

danmarsden:
yes - that looks better to me - feel free to send through a new PR with the other changes requested and I'll try to review when I get some spare time.

mgmodell:
I think this meets your request, @danmarsden , but please let me know if it does not.
I did not include tests because I was struggling to get any tests to run. I really don't know phpunit or behat or even composer. It's been ages since I've touched php at all, to be honest.

danmarsden:
yeah - pretty close - if you look at the "files changed" tab here in github we only want to see files changed that are part of this fix. you patch inscludes changes to .gitignore, composer.json and composer.lock that shouldn't be there.

I'd also be inclined to stop using lang strings and just inject the course fullname and sessiondate eg:
echo html_writer::tag('h2', userdate($session->sessdate));

also - when printing $course->fullname we need to pass it through format_string in case it has any multi-lang tags etc.
eg:
echo html_writer::tag('h3', format_string($course->fullname, true, array('context' => $context)));

mgmodell:
How's this, @danmarsden ?

danmarsden:
yeah - looks good - I'm just not sure why it hasn't triggered the github actions to run codechecker etc in the PR... There's a small conflict in the version.php file but I can fix that manually - just need to work out why the actions aren't being triggered.

mgmodell:
Fantastic!
Any insight into when I might expect this to be released? Specifically, when my students will be able to benefit from it?

mgmodell:
@danmarsden now that that PR is in, can we get this in? Is there any chance this might be available before the fall semester begins?

danmarsden:
finally got round to this one - I've dropped it into the 5.0 and 5.1 supported branches. I'll roll out updates to the moodle marketplace with this release soon.

PR #908, opened 2026-05-26
Add Moodle Playground preview blueprint and PR action

erseco:
## About this PR — Moodle Playground

[Moodle Playground](https://moodle-playground.com) (source: [ateeducacion/moodle-playground](https://github.com/ateeducacion/moodle-playground)) is a project that runs a full Moodle site **entirely inside the browser** (PHP + SQLite compiled to WebAssembly), with no server, no Docker and no local install. Given a `blueprint.json` it provisions a fresh Moodle instance, installs plugins, creates users / courses / sample content, and opens on a chosen landing page.

This PR wires that up for `mod_attendance` so contributors and reviewers can **try and test the plugin in a real Moodle instance with one click** — useful both for users who just want to evaluate it and for reviewers who want to validate a change without standing up a local Moodle.

> ℹ️ **Live PR-preview demo:** see the same change applied on my fork at https://github.com/erseco/moodle-mod_attendance/pull/1 — once the workflow finishes, that PR's description gets a "Preview in Moodle Playground" button appended automatically. That is exactly what every future PR opened against this repo will look like once this is merged.

## Summary
- Adds `blueprint.json` describing a Moodle 5.1 demo site that installs `mod_attendance` from this branch (`installMoodlePlugin`), creates a demo course (`ATTDEMO01`) and preloads a sample attendance activity via a `runPhpCode` step that calls `attendance_add_instance()` directly (same pattern used in [`projectestac/moodle-mod_geogebra`](https://github.com/projectestac/moodle-mod_geogebra)'s blueprint). `mod_attendance` is fully self-contained — no external API or workaround required.
- Adds `.github/workflows/pr-playground-preview.yml` which uses [`ateeducacion/action-moodle-playground-pr-preview`](https://github.com/ateeducacion/action-moodle-playground-pr-preview) to publish a live Moodle Playground preview link on every same-repo PR.
- Adds a "Try in Moodle Playground" section + badge to `README.md` linking to the blueprint.

### DIFF
--- a/.github/workflows/pr-playground-preview.yml
+++ b/.github/workflows/pr-playground-preview.yml
@@ -1,0 +1,30 @@
+---
+name: PR Playground Preview
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+concurrency:
+  group: playground-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  playground-preview:
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ateeducacion/action-moodle-playground-pr-preview@main
+        with:
+          blueprint-file: blueprint.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          moodle-version: '5.1'
--- a/README.md
+++ b/README.md
@@ -1,5 +1,13 @@
 # ABOUT
 
+## Try in Moodle Playground
+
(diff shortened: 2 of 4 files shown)


danmarsden:
Hi @erseco! - sounds like an interesting project and I can definitely see how it would be useful for some people.

This isn't something I'm currently keen to pull into the repository for ongoing maintenance at this stage and  I'm not a big fan on the new github action to trigger this. 

I also have a small concern around the potential for future trademark issues - you should definitely review https://moodle.com/trademarks/ and possibly have a quick conversation with Moodle HQ Legal before you continue  ongoing use of the term "Moodle Playground" in your domain name - it may limit what you're able to do in future.

However, if you manage to get Moodle HQ's endorsement of the project and approach I'd be happy to reconsider this again in future.

erseco:
Hi @danmarsden,

Many thanks for the feedback, you are completely right.

I have just contacted Moodle HQ to ask for guidance on this, both on whether they would consider authorising the use of the term `moodle-playground`, or whether they would prefer the project to be renamed.

As soon as I receive a response from them, I’ll let you know.

Thanks again for pointing this out.

danmarsden:
no worries!

Issue #645, opened 2022-12-19
Allowing Multilangs for Default Status


RahelPhasen:
Hi,

we wanted to use Multilangs for the default status. Unfortunantely this is not possible at this point because the database schema doesn't allow more than 2 letters for the acronym and not more than 30 for the description. To solve this problem an increase would be necessary.

danmarsden:
Thanks for the suggestion - while I agree we should definitely support multi-lang in the status description field, I'm not quite sure about supporting this in the acronym field - that would need some further thought and a closer look at where the acronym field is being displayed and how it's used in the code.

If you have internal development capabilities, feel free to send through a pull request for consideration.

maehne:
At Bern University of Applied Sciences (BFH, Switzerland), we would also be interested in full multilang support for this activity, as we teach courses in German, French, and English. Certain courses are taught bilingual. It would be nice if that could also cover the acronyms, as they differ between languages.

danmarsden:
Thanks @maehne feel free to help by sending through a pull request or reach out privately if you would like to fund improvements.

PR #913, opened 2026-08-05
General fixes and improvements

vladaskidanovas:


danmarsden:
Thanks Vlad - can you please remove the s() functions on the description field? - the description field can contain html and longer term should really be supporting multi-lang as per #645 so we really need to be changing those to format_string calls - but that would need to be a separate patch along with testing around multi-lang etc.

